PR #1, opened 2016-05-31
Initial Commit: Implement Install Resource

alecxvs:
This is the initial implementation of the resource cookbook, including the install resource for installing and uninstalling Splunk and appropriate tests.

Most potential issues may be in formatting, style, and documentation.

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -1,0 +1,34 @@
+---
+driver:
+  name: vagrant
+
+provisioner:
+  name: chef_zero
+  require_chef_omnibus: 12.9
+
+platforms:
+  - name: centos-7.2
+  - name: debian-8.4
+  - name: windows
+    driver:
+      box: opentable/win-2012r2-standard-amd64-nocm
+  - name: centos-7.2-compat
+    provisioner:
+      require_chef_omnibus: 12.4
+    driver:
+      box: bento/centos-7.2
+
+suites:
+  - name: install
+    run_list:
+      - recipe[cerner_splunk_ingredient_test::splunk]
+      - recipe[cerner_splunk_ingredient_test::universal_forwarder]
+  - name: uninstall
+    run_list:
+      - recipe[cerner_splunk_ingredient_test::splunk]
+      - recipe[cerner_splunk_ingredient_test::universal_forwarder]
+      - recipe[cerner_splunk_ingredient_test::uninstall_splunk]
+      - recipe[cerner_splunk_ingredient_test::uninstall_universal_forwarder]
+
+verifier:
+  name: inspec
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,13 @@
+branches:
+  except:
+    - master
+language: ruby
+cache: bundler
+sudo: false
+bundler_args: --without local
+rvm:
+- 2.1.8
+script:
+- bundle exec rubocop
+- bundle exec foodcritic -f any .
+- bundle exec rspec test/unit
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,32 @@
+# Feature Development:
+Developing for cerner_splunk_ingredient should follow these guidelines for submitting issues and pull requests.
+
+### Submitting Issues
+Issues pertaining to cerner_splunk_ingredient should be submitted to Github's issue tracker.
+1. Issues will be triaged and marked with an appropriate label (e.g. bug, enhancement, question).
+  - A corresponding internal JIRA may be logged if there is a desire for internal discussion or code review.
+- The issue should be worked in a branch outside of the Cerner repo.
+- The issue will be marked for the next milestone following a pull request.
+
+### Pull Requests
+Pull requests for changes should be compared against the 'stable' branch.
+1. Your branch should use as few logical commits as possible with descriptive commit messages.
+  - _Note: these commits should **not** reference any issues, that will be done in the merge commit._
+- Your pull request should reference the issue it is related to.
+- The pull request is collectively reviewed and commented on, and must get two or more +1's, from reviewers and at least one maintainer of the repo.
+  - Any raised comments must be addressed and will be re-reviewed.
+- Once the pull request is approved, the changes will be merged by a maintainer.
+    - This is done manually, not by using the Github button, so that the fix version of metadata.rb can be incremented in the merge. ([Example](https://github.com/cerner/cerner_splunk/issues/41#issuecomment-70569000))
+    - The merge commit text should reference the issue and the pull request so that both are closed when the merge is pushed.
+
+# Release Process:
+1. An issue is made to release a new version
+    - The issue will summarize what was changed since the last milestone, and propose the commit to tag and its description.
+    - The issue will be labeled "release"
+- The issue is reviewed, and must get two or more +1s from reviewers and maintainers.
+- An annotated tag is created with the reviewed release text.
+    - Master is set to the annotated tag.
+- The cookbook is released to Supermarket
+- Cleanup
+    - The issue and existing milestone is closed
+    - A new milestone is created for the next version
--- a/Gemfile
+++ b/Gemfile
@@ -1,0 +1,20 @@
+source 'https://rubygems.org'
+
+gem 'berkshelf', '~> 4.3'
+gem 'chef', '~> 12.4'
+gem 'compat_resource', '~> 12.5', '>= 12.5.6'
+gem 'rake'
+gem 'rubocop'
+gem 'foodcritic'
+gem 'chefspec'
+
+group :local do
+  gem 'test-kitchen'
+  gem 'kitchen-vagrant'
+  gem 'kitchen-inspec'
+  gem 'winrm', '~> 1.8'
+  gem 'winrm-fs', '~> 0.4'
+  gem 'guard'
+  gem 'guard-rubocop'
+  gem 'guard-foodcritic'
+end
--- a/README.md
+++ b/README.md
@@ -1,0 +1,76 @@
+cerner_splunk_ingredient Cookbook
+=================================
+Resource cookbook which provides custom resources for installing Splunk.
+
+These resources can:
+- Install Splunk via downloadable package or archive
+
+Requirements
+------------
+Chef >= 12.4
+Ruby >= 2.1.8
+
+Supports Linux and Windows based systems with package support for Debian, Redhat,
+and Windows.
+
+Using resources of this cookbook to install and run Splunk means you agree to Splunk's EULA packaged with the software,
+also available online at http://www.splunk.com/en_us/legal/splunk-software-license-agreement.html
+
+---
+
+Resources
+---------
+
+### splunk_install
+Manages an installation of Splunk
+
+##### Action *:install*
+Installs Splunk or Universal Forwarder.
+
+Properties:
+
+| Name     |               Type(s)               | Required | Default                                  | Description                                                                                                                                                                                                                                                            |
+|:---------|:-----------------------------------:|:--------:|:-----------------------------------------|:-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| package  | `:splunk` or `:universal_forwarder` | **Yes**  |                                          | Specifies the Splunk package to install. You must specify the package, or name the resource for the package; for example, `package :splunk` or `splunk_install 'universal_forwarder' do ... end`.                                                                      |
+| version  |               String                | **Yes**  |                                          | Version of Splunk to install                                                                                                                                                                                                                                           |
+| build    |               String                | **Yes**  |                                          | Build number of the version                                                                                                                                                                                                                                            |
+| base_url |               String                |    No    | `'https://download.splunk.com/products'` | Base url to pull Splunk packages from. Use this if you are mirroring the downloads for Splunk packages. The resource will append the version, os, and filename to the url like so: `{base_url}/splunk/releases/0.0.0/linux/splunk-0.0.0-a1b2c3d4e5f6-Linux-x86_64.tgz` |
+
+##### Action *:uninstall*
+Removes Splunk or Universal Forwarder and all its configuration.
+
+Properties:
+
+| Name    |               Type(s)               | Required | Default | Description                                                                                                                                                                                         |
+|:--------|:-----------------------------------:|:--------:|:--------|:----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| package | `:splunk` or `:universal_forwarder` | **Yes**  |         | Specifies the Splunk package to uninstall. You must specify the package, or name the resource for the package; for example, `package :splunk` or `splunk_install 'universal_forwarder' do ... end`. |
+
+
+---
+
+Contributing
+------------
+
+Check out the [Github Guide to Contributing](https://guides.github.com/activities/contributing-to-open-source/)
+for some basic tips on contributing to open source projects, and make sure to read our [Contributing Guidelines](CONTRIBUTING)
+before submitting an issue or pull request.
+
+License and Authors
+-------------------
+- Author:: Alec Sears (alec.sears@cerner.com)
+
+```text
+Copyright:: 2016, Cerner Innovation, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+```
--- a/libraries/path_helpers.rb
+++ b/libraries/path_helpers.rb
@@ -1,0 +1,38 @@
+module CernerSplunk
+  # Helper methods for Splunk-related file and URL paths
+  module PathHelpers
+    # Provides constant defaults for Splunk's various default install directories based on platform and package.
+    # Returned hash is nested by package and os. For example, `default_install_dirs[:splunk][:linux]` returns
+    # the default directory for Splunk on Linux.
+    #
+    # @return [Hash] an index of default install directories for Splunk and the Universal Forwarder
+    def default_install_dirs
+      {
+        splunk: {
+          linux: '/opt/splunk',
+          windows: 'c:\Program Files\Splunk'
+        },
+        universal_forwarder: {
+          linux: '/opt/splunkforwarder',
+          windows: 'c:\Program Files\SplunkUniversalForwarder'
+        }
+      }
+    end
+
+    # Utility method for splitting the filename from a URL.
+    #
+    # @param url [String]
+    # @return [String] the filename at the end of the URL
+    def filename_from_url(url)
+      Pathname.new(URI.parse(url).path).basename.to_s
+    end
+
+    # Append the relative path of Splunk's version file to the home directory path
+    #
+    # @param splunk_home [String] the absolute path to Splunk's home directory
+    # @return [Pathname] the absolute path to Splunk's version file
+    def version_pathname(splunk_home)
+      Pathname.new(splunk_home).join('etc/splunk.version')
+    end
+  end
+end
--- a/metadata.rb
+++ b/metadata.rb
@@ -1,0 +1,17 @@
+name             'cerner_splunk_ingredient'
+maintainer       'Cerner Innovation, Inc.'
+maintainer_email 'splunk@cerner.com'
+license          'Apache 2.0'
+description      'Installs Splunk'
+long_description IO.read(File.join(File.dirname(__FILE__), 'README.md'))
+version          '0.1.0'
+
+source_url       'https://github.com/cerner/cerner_splunk_ingredient'
+issues_url       'https://github.com/cerner/cerner_splunk_ingredient/issues'
+
+supports         'redhat', '>= 5.5'
+supports         'ubuntu', '>= 12.04'
+supports         'windows', '>= 6.1'
+
+depends          'tar', '~> 0.7'
+depends          'compat_resource' # Chef <12.5 Compatability
--- a/resources/splunk_install.rb
+++ b/resources/splunk_install.rb
@@ -1,0 +1,212 @@
+# Cookbook Name:: cerner_splunk_ingredient
+# Resource:: splunk_install
+#
+# Resource for managing the installation of Splunk
+class SplunkInstall < ChefCompat::Resource
+  include CernerSplunk::PlatformHelpers, CernerSplunk::PathHelpers
+
+  resource_name :splunk_install
+
+  property :name, String, name_property: true, identity: true
+  property :package, [:splunk, :universal_forwarder], required: true
+  property :version, String, required: true
+  property :build, String, required: true
+  property :base_url, String, default: 'https://download.splunk.com/products'
+
+  default_action :install
+
+  def after_created
+    node.run_state['splunk_ingredient'] ||= { 'installations' => {} }
+
+    case name.downcase
+    when 'splunk' then package :splunk
+    when 'universal_forwarder' then package :universal_forwarder
+    else raise 'Package must be specified (:splunk or :universal_forwarder)'
+    end unless property_is_set? :package
+  end
+
+  ### Inherited Methods
+
+  # Must be overridden by a platform provider
+  def kernel_string
+    raise 'No implementation for current platform'
+  end
+
+  def package_name
+    @package_name ||= package_names[package][node['os'].to_sym]
+    raise "Unsupported Combination: #{package} + #{node['os']}" unless @package_name
+    @package_name
+  end
+
+  def install_dir
+    @install_dir ||= default_install_dirs[package][node['os'].to_sym]
+    raise "Unsupported Combination: #{package} + #{node['os']}" unless @install_dir
+    @install_dir
+  end
+
+  def package_url
+    @package_url ||=
+      case package
+      when :splunk then "#{base_url}/splunk/releases/#{version}/#{node['os']}/splunk"
+      when :universal_forwarder then "#{base_url}/universalforwarder/releases/#{version}/#{node['os']}/splunkforwarder"
+      end + "-#{version}-#{build}-#{kernel_string}"
+  end
+
+  def package_path
+    @package_path ||= Pathname.new(Chef::Config['file_cache_path']) + filename_from_url(package_url)
+  end
+
+  ### Inherited Actions
+
+  load_current_value do |desired|
+    package desired.package
+    version_file = version_pathname(install_dir)
+    current_value_does_not_exist! unless version_file.exist?
+
+    version_data = version_file.read
+    installed_version, installed_build = version_data.match(/VERSION=(\d(?:\.\d)+).+BUILD=([\w]+)/m).captures
+    version installed_version
+    build installed_build
+  end
+
+  action :install do
+    install_state = node.run_state['splunk_ingredient']
+    raise "Install at #{install_dir} already exists!" if install_state['installations'][install_dir]
+    install_state['current_installation'] = install_state['installations'][install_dir] = {
+      name: name,
+      package: package,
+      version: version,
+      build: build,
+      x64: x64_support
+    }
+
+    converge_if_changed :version, :build do
+      remote_file package_path.to_s do
+        source package_url
+        show_progress true if defined? show_progress # Chef 12.9 feature
+        notifies :delete, "remote_file[#{package_path}]", :delayed
+      end
+    end
+  end
+
+  action :uninstall do
+    install_state = node.run_state['splunk_ingredient']
+    install_state.delete('current_installation')
+    install_state['installations'].delete(install_dir)
+
+    directory install_dir do
+      recursive true
+      action :delete
+    end
+  end
+end
+
+###################################
+### Platform Specific Providers ###
+###################################
+# rubocop:disable Documentation
+
+class LinuxInstall < SplunkInstall
+  resource_name :splunk_install
+  provides :splunk_install, os: 'linux'
+
+  def kernel_string
+    x64_support ? 'Linux-x86_64.tgz' : 'Linux-i686.tgz'
+  end
+
+  action :install do
+    super()
+
+    converge_if_changed :version, :build do
+      tar_extract package_path do
+        action :extract_local
+        target_dir install_dir
+      end
+    end
+  end
+end
+
+class RedhatInstall < SplunkInstall
+  resource_name :splunk_install
+  provides :splunk_install, platform_family: 'rhel'
+
+  def kernel_string
+    x64_support ? 'linux-2.6-x86_64.rpm' : 'i386.rpm'
+  end
+
+  action :install do
+    super()
+
+    converge_if_changed :version, :build do
+      rpm_package package_name do
+        source package_path.to_s
+        action :install
+      end
+    end
+  end
+
+  action :uninstall do
+    rpm_package package_name do
+      action :remove
+    end
+
+    super()
+  end
+end
+
+class DebianInstall < SplunkInstall
+  resource_name :splunk_install
+  provides :splunk_install, platform_family: 'debian'
+
+  def kernel_string
+    x64_support ? 'linux-2.6-amd64.deb' : 'linux-2.6-intel.deb'
+  end
+
+  action :install do
+    super()
+
+    converge_if_changed :version, :build do
+      dpkg_package package_name do
+        source package_path.to_s
+        action :install
+      end
+    end
+  end
+
+  action :uninstall do
+    dpkg_package package_name do
+      action :purge
+    end
+
+    super()
+  end
+end
+
+class WindowsInstall < SplunkInstall
+  resource_name :splunk_install
+  provides :splunk_install, os: 'windows'
+
+  def kernel_string
+    x64_support ? 'x64-release.msi' : 'x86-release.msi'
+  end
+
+  action :install do
+    super()
+
+    converge_if_changed :version, :build do
+      windows_package package_name do
+        source package_path.to_s
+        action :install
+        options 'LAUNCHSPLUNK=0 INSTALL_SHORTCUT=0 AGREETOLICENSE=Yes'
+      end
+    end
+  end
+
+  action :uninstall do
+    windows_package package_name do
+      action :remove
+    end
+
+    super()
+  end
+end
--- a/test/integration/install/spec.rb
+++ b/test/integration/install/spec.rb
@@ -1,0 +1,31 @@
+windows = os.windows?
+
+control 'basic_install' do
+  impact 0.4
+  title 'Install Splunk'
+  tag 'splunk_install'
+
+  describe package(windows ? 'Splunk Enterprise' : 'splunk') do
+    it { is_expected.to be_installed }
+    its('version') { is_expected.to match(/6\.3\.4(\.0)?(-cae2458f4aef)?/) }
+  end
+
+  describe command('ps -eo comm= | grep splunk') do
+    its('exit_status') { is_expected.to eq 1 }
+  end
+end
+
+control 'forwarder_install' do
+  impact 0.4
+  title 'Install Universal Forwarder'
+  tag 'splunk_install'
+
+  describe package(windows ? 'UniversalForwarder' : 'splunkforwarder') do
+    it { is_expected.to be_installed }
+    its('version') { is_expected.to match(/6\.3\.4(\.0)?(-cae2458f4aef)?/) }
+  end
+
+  describe command('ps -eo comm= | grep splunk') do
+    its('exit_status') { is_expected.to eq 1 }
+  end
+end
--- a/test/integration/uninstall/spec.rb
+++ b/test/integration/uninstall/spec.rb
@@ -1,0 +1,27 @@
+control 'basic_uninstall' do
+  impact 0.4
+  title 'Uninstall Splunk'
+  tag 'splunk_install'
+
+  describe package(os.windows? ? 'Splunk Enterprise' : 'splunk') do
+    it { is_expected.not_to be_installed }
+  end
+
+  describe command('ps -eo comm= | grep splunk') do
+    its('exit_status') { is_expected.to eq 1 }
+  end
+end
+
+control 'forwarder_uninstall' do
+  impact 0.4
+  title 'Uninstall Universal Forwarder'
+  tag 'splunk_install'
+
+  describe package(os.windows? ? 'UniversalForwarder' : 'splunkforwarder') do
+    it { is_expected.not_to be_installed }
+  end
+
+  describe command('ps -eo comm= | grep splunk') do
+    its('exit_status') { is_expected.to eq 1 }
+  end
+end
--- a/test/unit/spec/default_spec.rb
+++ b/test/unit/spec/default_spec.rb
@@ -1,0 +1,119 @@
+require_relative 'spec_helper'
+require_relative 'shared_examples'
+include CernerSplunk::PathHelpers, CernerSplunk::PlatformHelpers
+
+describe 'splunk_install' do
+  let(:runner_params) { { platform: 'redhat', version: '7.1' } }
+  let(:test_params) { { name: 'splunk', build: 'cae2458f4aef', version: '6.3.4' } }
+
+  let(:windows_opts) { 'LAUNCHSPLUNK=0 INSTALL_SHORTCUT=0 AGREETOLICENSE=Yes' }
+
+  let(:mock_run_state) { {} }
+  let(:chef_run) do
+    ChefSpec::SoloRunner.new({ step_into: ['splunk_install'] }.merge!(runner_params)) do |node|
+      node.set['test_parameters'] = test_params
+      node.set['run_state'].merge!(mock_run_state)
+    end.converge('cerner_splunk_ingredient_test::unit_test')
+  end
+
+  let(:run_state) { chef_run.node.run_state['splunk_ingredient'] }
+
+  describe 'action :install' do
+    platform_package_matrix.each do |platform, versions|
+      versions.each do |version, packages|
+        describe "on #{platform} #{version}" do
+          packages.each do |package, expected_url|
+            include_examples 'standard install', platform, version, package, expected_url
+          end
+        end
+      end
+    end
+
+    context 'with a non-package name' do
+      let(:test_params) { { name: 'Logmaster', package: :universal_forwarder, build: 'cae2458f4aef', version: '6.3.4' } }
+
+      it 'should install' do
+        expect(run_state['current_installation']).to eq(
+          name: 'Logmaster',
+          package: :universal_forwarder,
+          version: '6.3.4',
+          build: 'cae2458f4aef',
+          x64: true
+        )
+
+        expect(chef_run).to install_rpm_package('splunkforwarder')
+      end
+    end
+
+    context 'with base_url' do
+      let(:expected_url) do
+        'https://repo.internet.website/splunk/universalforwarder/releases/6.3.4/linux/splunkforwarder-6.3.4-cae2458f4aef-linux-2.6-x86_64.rpm'
+      end
+      let(:package_path) { "./test/unit/.cache/#{filename_from_url(expected_url)}" }
+      let(:test_params) do
+        {
+          package: :universal_forwarder,
+          build: 'cae2458f4aef',
+          version: '6.3.4',
+          base_url: 'https://repo.internet.website/splunk'
+        }
+      end
+
+      it 'should download the package' do
+        expect(chef_run).to create_remote_file(package_path).with(source: expected_url)
+      end
+    end
+
+    context 'when package is not specified' do
+      let(:test_params) { { name: 'hotcakes', build: 'cae2458f4aef', version: '6.3.4' } }
+
+      it 'should fail the Chef run' do
+        expect { chef_run }.to raise_error(/Package must be specified.*/)
+      end
+    end
+
+    context 'when version is not specified' do
+      let(:test_params) { { build: 'cae2458f4aef' } }
+
+      it 'should fail the Chef run' do
+        expect { chef_run }.to raise_error(Chef::Exceptions::ValidationFailed, /.* version is required/)
+      end
+    end
+
+    context 'when build is not specified' do
+      let(:test_params) { { version: '6.3.4' } }
+
+      it 'should fail the Chef run' do
+        expect { chef_run }.to raise_error(Chef::Exceptions::ValidationFailed, /.* build is required/)
+      end
+    end
+
+    context 'when platform is not supported' do
+      let(:runner_params) { { platform: 'aix', version: '7.1' } }
+      let(:test_params) { { name: 'splunk', build: 'cae2458f4aef', version: '6.3.4' } }
+
+      it 'should fail the Chef run' do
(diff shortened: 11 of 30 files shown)


test/integration/install/spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    its('exit_status') { is_expected.to eq 1 }
+  end
+  describe group('splunk') do
+    it { is_expected.to exist }
+  end unless windows
+  describe user('splunk') do
+    it { is_expected.to exist }
+    its('group') { is_expected.to eq 'splunk' } unless windows
+  end
+end
```
alecxvs: Inconsistent newlines

test/integration/install/spec.rb
```diff
@@ -0,0 +1,46 @@
+# Encoding: utf-8
```
alecxvs: Line after Encoding inconsistent in most of the files. There should probably be a new line here.

resources/splunk_install.rb
```diff
@@ -0,0 +1,220 @@
...
+  action :uninstall do
+    rpm_package package_name do
+      action :remove
+    end
+  end
+end
+
+class DebianInstall < SplunkInstall
+  resource_name :splunk_install
+  # use_automatic_resource_name
```
alecxvs: Remove

README.md
```diff
@@ -0,0 +1,68 @@
...
+Contributing
+------------
+
+Check out the [Github Guide to Contributing](https://guides.github.com/activities/contributing-to-open-source/)
+for some basic tips on contributing to open source projects, and make sure to read our [Contributing Guidelines](CONTRIBUTING)
+before submitting an issue or pull request.
+
+License and Authors
+-------------------
+- Author:: Alec Sears (alec.sears@cerner.com)
```
alecxvs: Add more people?

test/integration/install/spec.rb
alecxvs: dbadf53922e6dcc4b16f4829a189d31f84255b63

test/integration/install/spec.rb
alecxvs: dbadf53922e6dcc4b16f4829a189d31f84255b63

resources/splunk_install.rb
alecxvs: dbadf53922e6dcc4b16f4829a189d31f84255b63

acharlieh:
Let's break up commits in a more logical manner... like "Basic project structure" and "Install Resource"

.kitchen.yml
```diff
@@ -0,0 +1,33 @@
...
+  name: chef_zero
+  require_chef_omnibus: 12.9
+
+platforms:
+  - name: centos-7.2
+  - name: debian-8.4
+  - name: windows
+    driver:
+      box: win2012r2-standard-nocm-1.1.0
+      box_url: http://repo.release.cerner.corp/nexus/content/repositories/vagrant/com/cerner/vagrant/win2012r2-standard-nocm/1.1.0/win2012r2-standard-nocm-1.1.0.box
```
acharlieh: This isn't public, so we can't use it. instead what about using a box of opentable/win-2012r2-standard-amd64-nocm 

https://atlas.hashicorp.com/opentable/boxes/win-2012r2-standard-amd64-nocm/

CONTRIBUTING.md
```diff
@@ -0,0 +1,35 @@
...
+  - _Note: these commits should **not** reference any issues, that will be done in the merge commit._
+- Your pull request should reference the issue it is related to.
+- The pull request is collectively reviewed and commented on, and must get two or more +1's, from reviewers and at least one maintainer of the repo.
+  - Any raised comments must be addressed and will be re-reviewed.
+- Once the pull request is approved, the changes will be merged by a maintainer.
+    - This is done manually, not by using the Github button, so that the fix version of metadata.rb can be incremented in the merge. ([Example](https://github.com/cerner/cerner_splunk/issues/41#issuecomment-70569000))
+    - The merge commit text should reference the issue and the pull request so that both are closed when the merge is pushed.
+
+# Release Process:
+1. A pull request with the release commit is made to the 'stable' branch.
```
acharlieh: hm... we should update this text in the cerner_splunk one, because we no longer use a PR for releases since we increment version at merge of pull request time so release is just tagging. We use an issue proposing the tag sha, tag text, issue labeled release, and summarize changes.

README.md
```diff
@@ -0,0 +1,68 @@
...
+| service_user     |               String                |          No          | `'splunk'`                              | User to run the Splunk process                                                                                                                                                                                                                                           |
+| service_password |               String                |          No          |                                         | Shadow hash of the password for the user to run the Splunk process. The resource will attempt to install Splunk as a system user, but some platforms or configurations may require a password.                                                                           |
+| source_url       |               String                |          No          | `'http://download.splunk.com/products'` | Base url to pull Splunk packages from. Use this if you are mirroring the downloads for Splunk packages. The resource will append the version, os, and filename to the url like so: `{source_url}/splunk/releases/0.0.0/linux/splunk-0.0.0-a1b2c3d4e5f6-Linux-x86_64.tgz` |
```
acharlieh: We should default this to the https url.

README.md
acharlieh: You're the only author so far :) we'll add more as more people contribute significantly to it.

resources/splunk_install.rb
```diff
@@ -0,0 +1,221 @@
...
+
+  def after_created
+    # Validate Package
+    case name.downcase
+    when 'splunk' then package :splunk
+    when 'universal_forwarder' then package :universal_forwarder
+    else raise 'Package must be specified (:splunk or :universal_forwarder)'
+    end unless package
+
+    raise "Invalid package: #{package}" unless %i(splunk universal_forwarder).include? package
```
acharlieh: Wouldn't this be covered by the validation of the package property?

README.md
alecxvs: 6a1302f202a615c5d2a4d9fe7f110e9dbca57f3d

.kitchen.yml
alecxvs: 6a1302f202a615c5d2a4d9fe7f110e9dbca57f3d
bd7246d83652b4d8350e3a64431a7e0874cfe7f1

alecxvs:
Initial commit rebased into multiple:

Log message | Commit Hash
-------------------|-------------------
Initial Project Structure | 9b8c865
Implement splunk_install resource | f9b7eac
Unit Tests | 286a1e5
Integration Tests | cf63718

resources/splunk_install.rb
alecxvs: 4867bb4

CONTRIBUTING.md
alecxvs: It sounds like that should be the case here, too:
679ccab285a18216eb8f6aacbcefd4ed4fb69954

nmattam:
Looks like your email is not linked to the commits.

alecxvs:
Fixed commit author

metadata.rb
```diff
@@ -0,0 +1,20 @@
...
+version          '1.0.0'
+
+source_url       'https://github.com/cerner/cerner_splunk_ingredient'
+issues_url       'https://github.com/cerner/cerner_splunk_ingredient/issues'
+
+supports         'redhat', '>= 5.5'
+supports         'ubuntu', '>= 12.04'
+supports         'windows', '>= 6.1'
+
+# depends          'poise-service', '~> 1.3.0' #Will use this for future service resource
```
acharlieh: I would remove it for right now then. No need to have an extra comment here.

alecxvs: f4ec2dbe9b3040fe83d6d2143797c9c9555684ca

libraries/path_helpers.rb
```diff
@@ -0,0 +1,24 @@
+# Encoding: utf-8
+
+module PathHelpers
```
acharlieh: General comment... should we be name-spacing our classes and modules in a module named for the cookbook so that generic names like "PathHelpers" aren't accidentally reused or worse accidentally monkey patched elsewhere?

alecxvs: Yeah, that's probably a good idea. I was thinking of the modules organizationally at first, which that doesn't quite work now that I have two helper modules.

resources/splunk_install.rb
```diff
@@ -0,0 +1,219 @@
...
+
+  def package_name
+    case package
+    when :splunk then 'Splunk Enterprise'
+    when :universal_forwarder then 'UniversalForwarder'
+    end
+  end
+
+  # kernel_string = x64_support ? 'windows-64' : 'windows-32'
+  # extension = 'zip'
```
acharlieh: also comments from development that can be cleaned up.

resources/splunk_install.rb
```diff
@@ -0,0 +1,219 @@
...
+  def kernel_string
+    x64_support ? 'x64-release.msi' : 'x86-release.msi'
+  end
+
+  action :install do
+    raise 'Windows installation will fail unless license is accepted.' unless accept_license
```
acharlieh: Do we even want to provide the option to not programmatically accept the license then?

resources/splunk_install.rb
```diff
@@ -0,0 +1,219 @@
...
+
+    converge_if_changed :version, :build do
+      rpm_package package_name do
+        source package_path.to_s
+        action :install
+      end
+    end
+  end
+
+  action :uninstall do
```
acharlieh: Should you be calling super after invoking remove on the package resource?

resources/splunk_install.rb
```diff
@@ -0,0 +1,219 @@
...
+
+  resource_name :splunk_install
+
+  property :name, String, name_property: true, identity: true
+  property :package, [:splunk, :universal_forwarder], required: true
+  property :service_user, String, default: 'splunk'
+  property :service_password, String
+  property :version, String, required: true
+  property :build, String, required: true
+  property :source_url, String, default: 'https://download.splunk.com/products'
```
acharlieh: I might be being pedantic, but is this a source url or mirror url or a base url?

resources/splunk_install.rb
```diff
@@ -0,0 +1,219 @@
+# Encoding: utf-8
+
+class SplunkInstall < ChefCompat::Resource
+  include PlatformHelpers, PathHelpers
+
+  resource_name :splunk_install
+
+  property :name, String, name_property: true, identity: true
+  property :package, [:splunk, :universal_forwarder], required: true
+  property :service_user, String, default: 'splunk'
```
acharlieh: do we want to manage group as well? or no?

resources/splunk_install.rb
alecxvs: Fair point, it can be made required with a true value only since it's not used elsewhere
That makes me wonder if we should force license agreement before installing on any platform in the future (if that's even possible considering that it has to be done on startup on linux)

resources/splunk_install.rb
alecxvs: Is there a use for doing so? Having a system user assumes there's a group named after the user, so that's not going to change. We could allow adding extra groups, but consumers could also do that. It might avoid cloning the user resource, however.

resources/splunk_install.rb
alecxvs: It feels like it could be any of the three, but it might be leaning towards a base URL.

test/integration/install/spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+windows = os.windows?
+
+control 'basic_install' do
+  impact 0.4
+  title 'Install Splunk'
+  tag 'splunk_install'
+
+  describe package(os.windows? ? 'Splunk Enterprise' : 'splunk') do
```
acharlieh: nit... you defined windows as `os.windows?` above.... do you want to use it here? or is it not saving you enough that you should just use os.windows? on lines 21 and 25

test/unit/spec/default_spec.rb
```diff
@@ -0,0 +1,260 @@
...
+    }
+  },
+  suse: { # Generic Linux
+    '12.0': {
+      splunk: 'https://download.splunk.com/products/splunk/releases/6.3.4/linux/splunk-6.3.4-cae2458f4aef-Linux-x86_64.tgz',
+      universal_forwarder: 'https://download.splunk.com/products/universalforwarder/releases/6.3.4/linux/splunkforwarder-6.3.4-cae2458f4aef-Linux-x86_64.tgz'
+    }
+  }
+}
+describe 'splunk_install' do
```
acharlieh: Nit: probably want a newline between your hash definition and your spec.

.travis.yml
```diff
@@ -0,0 +1,12 @@
+branches:
+  except:
+    - master
+language: ruby
+cache: bundler
+sudo: false
+rvm:
+- 2.1.6
```
acharlieh: So you have 2.1.6 here... current version of chefdk ships 2.1.8... but your Gemfile doesn't install on either of these: 

    ...
    Using ruby_dep 1.3.1
    Installing listen 3.1.5

    Gem::InstallError: listen requires Ruby version >= 2.2.3, ~> 2.2.
    An error occurred while installing listen (3.1.5), and Bundler cannot continue.
    Make sure that `gem install listen -v '3.1.5'` succeeds before bundling.

README.md
```diff
@@ -0,0 +1,68 @@
...
+Supports Linux and Windows based systems with package support for Debian, Redhat,
+and Windows.
+
+---
+
+Resources
+---------
+
+### splunk_install
+Installs Splunk. Easy.
```
acharlieh: "Use the splunk_install resource to manage the core parts of a Splunk installation" perhaps? The resource itself doesn't install splunk, using one of the actions you take on the resource asserts that the desired version of Splunk is installed (subject to package manager constraints of course). (Also I'm not sure if I'm crazy or not with this idea, but I wonder if like with ReST, Resources should be Nouns, and you perform Action (verbs) upon them.)

metadata.rb
```diff
@@ -0,0 +1,19 @@
+# Encoding: utf-8
+
+name             'cerner_splunk_ingredient'
+maintainer       'Cerner Innovation, Inc.'
+maintainer_email 'splunk@cerner.com'
+license          'Apache 2.0'
+description      'Installs Splunk'
+long_description IO.read(File.join(File.dirname(__FILE__), 'README.md'))
+version          '1.0.0'
```
acharlieh: Something to think about, do we want to start at 1.0.0 or do we want to start with 0.1.0 ?

resources/splunk_install.rb
```diff
@@ -0,0 +1,219 @@
...
+    installed_version, installed_build = version_data.match(/VERSION=(\d(?:\.\d)+).+BUILD=([\w]+)/m).captures
+    version installed_version
+    build installed_build
+  end
+
+  action :install do
+    state = node.run_state['splunk_ingredient'] ||= {}
+    state['packages'] ||= {}
+
+    raise "Installation resource with name #{name} already exists!" if state['packages'][name]
```
acharlieh: should we be keying by name? or should we key by install_dir ?

resources/splunk_install.rb
```diff
@@ -0,0 +1,219 @@
...
+    state['current_package'] = state['packages'][name] = {
+      package: package,
+      version: version,
+      build: build,
+      x64: x64_support
```
acharlieh: I'm not sure we need this, as this comes simply from an ohai attribute yes?

resources/splunk_install.rb
acharlieh: We probably should be removing information from run state in the case of an uninstallation as well?

resources/splunk_install.rb
acharlieh: Hm...good point... Do we do anything with the user in this resource other than ensure it's created? Do we need it for install time on Windows (I don't think I see it being used)? Or could user / service management for an installation be a different resource or different action? (action :enable / :disable to manage the service?)

metadata.rb
alecxvs: If we're going to be making changes parallel to cerner_splunk, 0.1.0 is a better choice.

resources/splunk_install.rb
alecxvs: We don't seem to be using the user at all. Windows doesn't require it, but it needs to be set at install time I think. User creation should be part of service setup (which is intertwined with the Windows installation process but we can deal with that when we get to it).

I'm probably going to remove user creation then and it will be part of the service resource.

resources/splunk_install.rb
alecxvs: It's not as clear since it could be represented as x86_64 or amd64 (or other combinations of weirdness on other platforms that aren't supported). Though we don't really need to provide this in the run state, I'm already producing a boolean for 64-bit support and it doesn't hurt to throw that in with the package state (plus it makes it clear we're installing the 64-bit version of Splunk and is more easily testable).

resources/splunk_install.rb
alecxvs: We should key by install_dir. I forgot I moved this check to install, and I wanted to keep the name so I could easily test uninstall.

We should probably store installations in run_state instead of packages, as that's what the splunk_install resource deals with. I'm curious now, though, if I could/should load existing installations into run_state at compile time.

README.md
alecxvs: I wanted to go the route of nouns and verbs, but opted for splunk_install as it was probably going to be a huge resource to cram configure in with it. Chances are, though, that the resource-side of things isn't going to be terribly complicated. I fully intend to rename the resource to just 'splunk' if I add configuration and it's clean enough.

test/integration/install/spec.rb
alecxvs: `windows` should be used.
Why I'm assigning `windows` instead of just using `os.windows?` isn't to save space, but because `os` isn't defined specifically when I use it here: `describe ... end unless os.windows?`
See: https://github.com/chef/inspec/issues/628

metadata.rb
alecxvs: f01e16056a3dbabbd9315055660a06bebd8689e7

.travis.yml
alecxvs: f01e16056a3dbabbd9315055660a06bebd8689e7

Should it match ChefDK version?

test/unit/spec/default_spec.rb
alecxvs: f01e16056a3dbabbd9315055660a06bebd8689e7

test/integration/install/spec.rb
alecxvs: f01e16056a3dbabbd9315055660a06bebd8689e7

resources/splunk_install.rb
alecxvs: If we want to be super sure the splunk directory gets deleted, then yes. Package removal should be deleting the directory (or most of it, I think the rpm removal leaves a single redhat specific script)

libraries/path_helpers.rb
alecxvs: f870316e401a5b4fb7cd81171497d0bebe208615

resources/splunk_install.rb
alecxvs: f01e160

.travis.yml
```diff
@@ -0,0 +1,12 @@
...
+    - master
+language: ruby
+cache: bundler
+sudo: false
+rvm:
+- 2.2.3
+script:
+- bundle exec rubocop
+- bundle exec foodcritic -f any .
+- bundle exec rspec test/unit
```
alecxvs: Should I add kitchen tests to this?

alecxvs: Also rubocop is being run on all the gems in the bundle, which is pointless.

resources/splunk_install.rb
alecxvs: aa54e36fa390277513a97ff3bc03d9dcc923ad59

resources/splunk_install.rb
alecxvs: c2e8a3bdf6cd33e9779c4cc76d0802e600738ca9

README.md
alecxvs: c2e8a3bdf6cd33e9779c4cc76d0802e600738ca9

resources/splunk_install.rb
alecxvs: c2e8a3bdf6cd33e9779c4cc76d0802e600738ca9

resources/splunk_install.rb
alecxvs: I ended up super()ing anyways to share run_state changes
c2e8a3bdf6cd33e9779c4cc76d0802e600738ca9

resources/splunk_install.rb
alecxvs: c2e8a3bdf6cd33e9779c4cc76d0802e600738ca9

.travis.yml
acharlieh: It'd be nice if we could do kitchen tests as well, we would need to figure out how to do kitchen tests on Travis (or another cloud based CI system) and ensure it's not brittle. Just not something we've done yet, but it'd be cool if we could.

alecxvs: Is there a way to run manual travis checks with different configs? I'd like to play with it a bit to try and get a chef-zero run and run inspec, if not kitchen as a whole (there's no way to run vagrant boxes on travis, so we would end up running chef locally and testing that either way)

alecxvs:
I added unit tests for removing run_state on uninstall, and changed how I was stubbing the version file in another test. Instead of explicitly calling the chef runner again, it now does the stubbing in a let that's called before the runner runs. I think it's still running for each individual example, but there's less mess in the code now and less potential for it to be running redundantly.

Also renamed state to install_state (in splunk_install.rb) because apparently state is something defined by the Chef Resource class.

9becf50fd55ab778bc0dbe49574c952a096e11c7

nmattam:
+1

libraries/path_helpers.rb
```diff
@@ -0,0 +1,40 @@
...
+    # @return [String] the filename at the end of the URL
+    def filename_from_url(url)
+      Pathname.new(URI.parse(url).path).basename.to_s
+    end
+
+    # Append the relative path of Splunk's version file to the home directory path
+    #
+    # @param splunk_home [String] the absolute path to Splunk's home directory
+    # @return [Pathname] the absolute path to Splunk's version file
+    def version_pathname(splunk_home = install_dir)
```
acharlieh: Where does this default value come from?

resources/splunk_install.rb
```diff
@@ -0,0 +1,223 @@
...
+        source package_url
+        show_progress true if defined? show_progress # Chef 12.9 feature
+        notifies :delete, "remote_file[#{package_path}]", :delayed
+      end
+    end
+  end
+
+  action :uninstall do
+    install_state = node.run_state['splunk_ingredient']
+    install_state.delete('current_installation')
```
acharlieh: Should you be removing this only if this resource corresponds to the current_installation?

libraries/path_helpers.rb
alecxvs: I'm not sure how to describe it in the documentation, but it's a little magic because that's a method defined in splunk_install where the resource is being mixed in. There's no reason to be doing that though since I call it maybe once, I can pass the parameter normally.

resources/splunk_install.rb
alecxvs: That's a curious question, because it's easier and more sanitary in my mind to remove the current_installation as that entry in run_state is meant for chaining future resources and performing actions immediately after an install. It would be very unusual to do so after an uninstall, and I don't think we need that logic right now. It's not an impossible scenario, though, either. Just maybe one that has some smelly code.

Furthermore, I intend to follow the pattern of setting current_installation in every resource that specifies an installation, so with uninstall not having an installation anymore to set this to, it makes more sense that there is no current_installation.

libraries/path_helpers.rb
alecxvs: 34547d81c56dd573c7e431b0ddb8d1939a0d07dd

Gemfile
```diff
@@ -0,0 +1,18 @@
+source 'https://rubygems.org'
+
+gem 'berkshelf', '> 3.1'
```
jdmurphy: Minor in the scope of things, but considering the issues we have run into with the 3.x series (Faraday gzip comes to mind), I would stick to the 4.x series of berkshelf.

README.md
```diff
@@ -0,0 +1,66 @@
...
+Supports Linux and Windows based systems with package support for Debian, Redhat,
+and Windows.
+
+---
+
+Resources
+---------
+
+### splunk_install
+Manages an installation of splunk
```
jdmurphy: Minor, capital S on Splunk.

resources/splunk_install.rb
```diff
@@ -0,0 +1,223 @@
...
+  include CernerSplunk::PlatformHelpers, CernerSplunk::PathHelpers
+
+  resource_name :splunk_install
+
+  property :name, String, name_property: true, identity: true
+  property :package, [:splunk, :universal_forwarder], required: true
+  property :version, String, required: true
+  property :build, String, required: true
+  property :base_url, String, default: 'https://download.splunk.com/products'
+  property :accept_license, [TrueClass], required: true
```
jdmurphy: Don't we need `[TrueClass, FalseClass]` here to be able to do either true or false?

test/unit/spec/default_spec.rb
```diff
@@ -0,0 +1,305 @@
...
+              expect(chef_run.node.run_state['splunk_ingredient']['current_installation']).to eq(
+                name: 'splunk',
+                package: :splunk,
+                version: '6.3.4',
+                build: 'cae2458f4aef',
+                x64: true
+              )
+            end
+
+            it { is_expected.to create_remote_file(package_dir).with(source: urls[:splunk]) }
```
jdmurphy: We are mixing and matching test formats all over which irks me; it seems to me that it would be cleaner, more readable and easier to update in the future if we stuck with one format. I personally think we should stick with the following format:
```
it 'should do something' do
  expect(something).to happen
end
```

test/unit/spec/default_spec.rb
```diff
@@ -0,0 +1,305 @@
...
+
+            case platform
```
jdmurphy: I can see this kind of block multiple places in the test file. We should be able to move this to a shared_example to help clean and DRY this file up some.

test/unit/spec/default_spec.rb
```diff
@@ -0,0 +1,305 @@
...
+                  is_expected.to install_windows_package('Splunk Enterprise').with(
+                    options: 'LAUNCHSPLUNK=0 INSTALL_SHORTCUT=0 AGREETOLICENSE=Yes'
+                  )
+                end
+              end
+            else it { is_expected.to extract_local_tar_extract(package_dir).with(target_dir: '/opt/splunk') }
+            end
+
+            context 'when Splunk is already installed' do
+              let(:do_before) do
```
jdmurphy: Whats the reasoning for using this instead of the regular 
```
before do
...
end
```

blocks?

resources/splunk_install.rb
alecxvs: We would throw an error when giving false anyways, might as well fail fast by failing the validation.

test/unit/spec/default_spec.rb
alecxvs: A before block would run before each test, but oddly not before the subject (the chef run). I need to be able to stub things before the chef run too, so this let is called before the chef run is converged.

test/unit/spec/default_spec.rb
alecxvs: i'm not sure how I would accomplish that. There are multiple case platform blocks, but each of them have different conditions, and I don't know how I would combine all of them and still reduce code.

resources/splunk_install.rb
jdmurphy: That seems dissonant though. In the README, we outline that true or false are valid options for this attribute, but here we fail with a validation error if false is actually used.

test/unit/spec/default_spec.rb
jdmurphy: Fine...personal preference. You like that style, I like the other. If we can't do all one style though, then it is the mix and match which I still think we need to fix.

test/unit/spec/default_spec.rb
jdmurphy: At the very least you can have a shared example for installation of packages that relies on platform, and any parameters used in the install block. I can see that kind of thing all over this file. In addition, you can have shared contexts for things such as your contexts around "with the Splunk package" and "with the Universal Forwarder package". There are lots of ways we can clean this up.

test/unit/spec/default_spec.rb
alecxvs: Unfortunately it's impossible to use one-liners throughout, so we'll have to use the standard format.

test/unit/spec/default_spec.rb
alecxvs: I'm working on combining the package contexts.

alecxvs: c75a9507b2767e7ce0df595009bcfe7de35ffcb4

test/unit/spec/default_spec.rb
alecxvs: c75a9507b2767e7ce0df595009bcfe7de35ffcb4

resources/splunk_install.rb
alecxvs: b44cde1ad7ecc81a3bd9c73085de042e587905ed

Gemfile
alecxvs: 0551e39585dea006cbcbf284a30e5bf1ddfbb5f9

README.md
alecxvs: 0551e39585dea006cbcbf284a30e5bf1ddfbb5f9

alecxvs:
Squashing commits ⚡

resources/splunk_install.rb
```diff
@@ -0,0 +1,215 @@
...
+    converge_if_changed :version, :build do
+      rpm_package package_name do
+        source package_path.to_s
+        action :install
+      end
+    end
+  end
+
+  action :uninstall do
+    super()
```
acharlieh: We should have super() be the last call in each uninstall... remove package then clean up directory leftovers.

README.md
```diff
@@ -0,0 +1,69 @@
...
+| build          |               String                |       **Yes**        |                                          | Build number of the version                                                                                                                                                                                                                                            |
+| base_url       |               String                |          No          | `'https://download.splunk.com/products'` | Base url to pull Splunk packages from. Use this if you are mirroring the downloads for Splunk packages. The resource will append the version, os, and filename to the url like so: `{base_url}/splunk/releases/0.0.0/linux/splunk-0.0.0-a1b2c3d4e5f6-Linux-x86_64.tgz` |
+| accept_license |                true                 | **Yes (on Windows)** |                                          | Set if you accept Splunk's license agreement on install. Currently only for the Windows Package, as it requires the license to be accepted at install time. On Splunk for Linux, it can be accepted on first run with `--accept-license`.                              |
```
acharlieh: I've been thinking about this a lot... It's a parameter, that's required to be set to true in a package designed for automation. What if we removed this parameter, and instead put something to the effect of "Using resources of this cookbook to install and run Splunk means you agree to Splunk's EULA packaged with the software, online at http://www.splunk.com/en_us/legal/splunk-software-license-agreement.html" Then we can get rid of the option, and the logic reacting to having different options for it.

README.md
```diff
@@ -0,0 +1,69 @@
...
+
+##### Action *:uninstall*
+Removes Splunk or Universal Forwarder and all its configuration.
+
+Properties: _None_
```
acharlieh: Realized this isn't quite right... you need the package name here.

acharlieh:
Few minor changes left, then 👍 from me.

README.md
jdmurphy: +1

resources/splunk_install.rb
```diff
@@ -0,0 +1,215 @@
...
+  property :package, [:splunk, :universal_forwarder], required: true
+  property :version, String, required: true
+  property :build, String, required: true
+  property :base_url, String, default: 'https://download.splunk.com/products'
+  property :accept_license, TrueClass, required: true
+
+  default_action :install
+
+  def after_created
+    node.run_state['splunk_ingredient'] ||= { 'installations' => {} }
```
acharlieh: we should separate the initialization of splunk_ingredient, and splunk_ingredient.installations run states... the thought being with future resources, what if someone uses a different resource that doesn't deal with installations first?

alecxvs: I wonder if that change would be better made later when there is actually another resource using the run state. It would be trivial to extract this into a helper method and remain passive later, and I can't determine the design for very future resources yet. It could be the case that the run state for every other resource exists under the context of an installation (because logically there might not be anything we're managing outside of installed Splunk).

acharlieh: Good call... I was definitely over thinking that one.

resources/splunk_install.rb
alecxvs: 3685b1a00572aee2694b768382963aae1f1999df

README.md
alecxvs: 3312faca73163e7382aef00a074d78cf5e16943a

README.md
alecxvs: 18fa620bef53ba127a1daa4e97f074a16ee451b1

README.md
```diff
@@ -0,0 +1,76 @@
...
+
+##### Action *:uninstall*
+Removes Splunk or Universal Forwarder and all its configuration.
+
+Properties:
+
+| Name    |               Type(s)               | Required | Default | Description                                                                                                                                                                                       |
+|:--------|:-----------------------------------:|:--------:|:--------|:--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| package | `:splunk` or `:universal_forwarder` | **Yes**  |         | Specifies the Splunk package to install. You must specify the package, or name the resource for the package; for example, `package :splunk` or `splunk_install 'universal_forwarder' do ... end`. |
```
gravesb: minor, "...the splunk package to uninstall"

alecxvs: 146f45c

test/unit/spec/default_spec.rb
gravesb: I'm not sure I follow your reasoning.  All the tests still pass if I switch this to a regular before do block.  Unless you accidentally broke something along the way during the review, the chef_run is currently being executed for every example, not just once per example group.

resources/splunk_install.rb
```diff
@@ -0,0 +1,214 @@
...
+
+    version_data = version_file.read
+    installed_version, installed_build = version_data.match(/VERSION=(\d(?:\.\d)+).+BUILD=([\w]+)/m).captures
+    version installed_version
+    build installed_build
+  end
+
+  action :install do
+    install_state = node.run_state['splunk_ingredient']
+    raise "Install at #{install_dir} already exists!" if install_state['installations'][install_dir]
```
gravesb: So would this only happen if a node's run list contained this install resource multiple times right? Or I suppose if the install action was notified multiple times during a run.

resources/splunk_install.rb
```diff
@@ -0,0 +1,214 @@
...
+  end
+
+  action :install do
+    super()
+
+    converge_if_changed :version, :build do
+      windows_package package_name do
+        source package_path.to_s
+        action :install
+        options 'LAUNCHSPLUNK=0 INSTALL_SHORTCUT=0 AGREETOLICENSE=Yes'
```
gravesb: So from the doc, install_shortcut only applies for the full splunk installation.  I'm not sure if it hurts anything to pass in additional flags but we could pretty easily flex the options based on the package attribute.

gravesb:
Since we only support ruby 2.x (chef 12.x +) you can drop all the encoding comments in the source files.  utf-8 is the default in ruby 2+.

.kitchen.yml
```diff
@@ -0,0 +1,32 @@
...
+      require_chef_omnibus: 12.4
+    driver:
+      box: bento/centos-7.2
+
+suites:
+  - name: install
+    run_list:
+      - recipe[cerner_splunk_ingredient_test::splunk]
+      - recipe[cerner_splunk_ingredient_test::universal_forwarder]
+  - name: uninstall
```
gravesb: We should have a third test kitchen suite to test uninstalling the universal forwarder.  We should rename this suite (and the uninstall recipe) to be more specific to the splunk package.  e.g. uninstall_splunk

resources/splunk_install.rb
alecxvs: Yes, executing the install action multiple times pointing to the same installation path (currently, basically the same package) will throw this error. You can run any other actions (like uninstall) without encountering this error, but attempting to install it twice will be an error.

It should be possible to run the install, uninstall, then install in the same chef run without encountering this error, but that's weird and isn't tested.

resources/splunk_install.rb
alecxvs: Integration tests succeed so it doesn't seem to adversely affect the installation. If we add more options specific to the packages in the future, we should flex it, but that's not really necessary here.

alecxvs:
I did not know that 👍

test/unit/spec/default_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+          packages.each do |package, _|
+            include_examples 'standard uninstall', platform, version, package
+          end
+        end
+      end
+    end
+
+    context 'when package is not specified' do
+      let(:test_params) { { name: 'hotcakes', action: :uninstall } }
+      it 'fails to install' do
```
gravesb: nit, fails to uninstall.  Or possibly reword to just 'fails the chef run' in which case you could do the same for all the other failure tests so they can all be safely copy pasted around.

resources/splunk_install.rb
```diff
@@ -0,0 +1,214 @@
...
+  ### Inherited Methods
+
+  # Must be overridden by a platform provider
+  def kernel_string
+    raise 'No implementation for current platform'
+  end
+
+  def package_name
+    @package_name ||= package_names[package][node['os'].to_sym]
+    raise "Unsupported Combination: #{package} + #{node['os']}" unless @package_name
```
gravesb: need chef specs for this raise error. same for the raise in install_dir below.

test/unit/spec/shared_examples.rb
```diff
@@ -0,0 +1,127 @@
...
+          options: 'LAUNCHSPLUNK=0 INSTALL_SHORTCUT=0 AGREETOLICENSE=Yes'
+        )
+      end
+    else
+      it 'should install from the tgz package' do
+        expect(chef_run).to extract_local_tar_extract(package_path).with(target_dir: install_dir)
+      end
+    end
+
+    context "when '#{package}' is already installed" do
```
gravesb: can you add a context for when the package is already installed but with a different version?

.kitchen.yml
alecxvs: 49878a781929b6afc2aaa898a1b457fb6b50a11f

I added a recipe for universal forwarder and added it to the uninstall suite; is there any reason to split the uninstall suite for the two packages? There is no difference test-wise, and I intended to cut down on test time by not setting up chef omnibus on ~20 VMs.

alecxvs:
9d56b606c536f7b5c80dd398b369360ceea5d6b7

Removed encoding comments

.kitchen.yml
gravesb: No, that's a good call.  I just didn't think about adding the recipes to the same suite.

test/integration/uninstall/spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  impact 0.4
+  title 'Uninstall Splunk'
+  tag 'splunk_install'
+
+  describe package(os.windows? ? 'Splunk Enterprise' : 'splunk') do
+    it { is_expected.not_to be_installed }
+  end
+
+  describe command('ps -eo comm= | grep splunk') do
+    its('exit_status') { should eq 1 }
```
jdmurphy: Just wondering, why the difference between install and uninstall for `is_expected.to eq 1` and `should eq 1`?

alecxvs: Copy paste error, probably

test/unit/spec/shared_examples.rb
alecxvs: 038f2995d0210267348539f99f18742c49cdaf9f

test/unit/spec/default_spec.rb
alecxvs: 2a95c28a6d482f92890db0aab7f5792c5210018d

test/integration/uninstall/spec.rb
alecxvs: 6d76bbef842b7e565fc79c86e180b5f359bd8a94

resources/splunk_install.rb
alecxvs: I can test for the error, but I don't think I can test these methods specifically.

alecxvs: 521ee580f823588b64ce4b075f5448d9da3b8ed0

gravesb: Yeah, that's what I meant.  A test that causes this line to be executed.

.travis.yml
```diff
@@ -0,0 +1,14 @@
+branches:
+  except:
+    - master
+language: ruby
+cache: bundler
+sudo: false
+rvm:
```
jdmurphy: Do we really need separate testing for these three versions of ruby? At most I could see 2.2.x vs 2.3.x, but even those should be remarkably similar.

gravesb: That's a good point, I think chef's embedded ruby is still 2.1.x.  I have chef-dk 0.13.21 that comes packaged with chef 12.9.41 and ruby 2.1.8. So I think this should definitely include ruby 2.1.<something>.

resources/splunk_install.rb
gravesb: Follow up question, which one of these raise statements does that test hit?  I think this one is dead code since the install_dir method is currently always called before this one (if I'm manually tracing the code paths correctly).

.travis.yml
alecxvs: Really? I expected they would be using a newer Ruby by now.
Regardless, Guard in the bundle requires >= 2.2.3 and since travis installs the whole gem bundle, I can't test 2.1.x. I will limit it to just 2.2.5.

4328e2192592966f6e3a692e192431207df046fc

acharlieh: I had missed your reply [to my previous comment](https://github.com/cerner/cerner_splunk_ingredient/pull/1#discussion_r65640587) but we should indeed be running tests under ruby 2.1.x lest we have some syntax or quirk that doesn't jive with the version of ruby that ships with Chef. 

Guard seems like a utility that you'd use only with local development... should it be in the bundle? Or should we have a way to exclude it for Travis builds? Also there's an interesting note on https://github.com/guard/guard saying that with the right version of bundler, it should install the appropriate version for your ruby... maybe that's something to look into as well?

jdmurphy: Couldn't we just move guard and all the related stuff to a development block and do bundle install without development?

alecxvs: Is that possible with travis? I don't specify the bundle install command, just that it should use bundler.

jdmurphy: See [this page](https://docs.travis-ci.com/user/languages/ruby). It describes a `bundler_args` option which sounds like exactly what we need.

alecxvs: Exactly what we need indeed. I can cut out kitchen gems too since we don't use those in travis.

I had no success installing guard on 2.1.8 or 2.1.10. They say bundler should be able to resolve it, but it doesn't. So specifying the args is necessary unless I remove the guard gems from the bundle entirely (seems weird to me given the Guardfile referencing Guard gems for RSpec and Foodcritic)

test/unit/spec/default_spec.rb
alecxvs: I need to look into that, because chef runs should not be per example.

.travis.yml
jdmurphy: Rockets haven't been necessary (barring having some characters in the key) since Ruby 2.0 I believe.

alecxvs:
Referencing this comment out here because it's buried in an outdated diff on a squashed commit

https://github.com/cerner/cerner_splunk_ingredient/pull/1#discussion_r66867234

> I'm not sure I follow your reasoning. All the tests still pass if I switch this to a regular before do block. Unless you accidentally broke something along the way during the review, the chef_run is currently being executed for every example, not just once per example group.

The main point is that apparently chef_run is converging for every example (because chef_run.converge is specified in a let). I'm not sure why before was not working for me previously, but it shouldn't be working for me now (and it is). Converging for every example is an absolute waste since ChefSpec only checks that resources were called and each example does not need to be isolated from each other. Furthermore, I wrote the contexts to separate differences in state between all the test cases.

I'm trying to devise a way of defining the chef_run for each example group, but there's no broader let. The closest thing I have is a subject, but I will have to copy paste it for every context and that doesn't solve my wasted computation time with the dozen other lets that are only specified and used once per context.
An effective solution would be collapsing all the examples in my context under one it block, but that makes for a terrible test output.

.travis.yml
alecxvs: I do have special characters in the key though, "7.2" for example. "7.2": "value" translates to a symbol key (and there is no way to specify a string key without a rocket, but I need to rewrite a lot of stuff to expect strings instead of symbols now anyways)

alecxvs: b2f678ca0339c5f41bdefa387dafa2d64d320d3a

alecxvs:
I'm going to switch most of the contexts so that expectations under a single chef run all reside under the same example. The output won't be great, but it will make more sense as far as determining when chef runs happen during unit tests, and it will speed up unit tests tremendously.

The alternative, being to only converge in example groups, ends up being terribly complex.

resources/splunk_install.rb
alecxvs: Yeah, that will be called first in any case.

gravesb:
Why PR into a stable branch instead of just using master (and making master only contain stable code)?

README.md
```diff
@@ -0,0 +1,76 @@
...
+=================================
+Resource cookbook which provides custom resources for installing Splunk.
+
+These resources can:
+- Install Splunk via downloadable package or archive
+
+Requirements
+------------
+Chef >= 12.4
+Ruby >= 2.2.3
```
gravesb: need to adjust this after our ruby version discussion earlier on this PR. Since chef 12.4.3 packages ruby 2.1.<something> we can't say we require 2.2.3 or above.

acharlieh:
@gravesb With regards to the `stable` versus `master` I figured we could start with following the same pattern as [cerner_splunk](https://github.com/cerner/cerner_splunk)... that said there could be an argument to be made to switch branches about in both repos.

alecxvs:
I don't see the practical purpose of having stable and master with master sitting at a tagged version. The tagged versions are suitable for getting releases, and if we really want to have a branch consistently pointing to the latest release, a branch named `latest` would be more concise.

README.md
alecxvs: 1f9d5e96d28047ec711c78d7f9a1f74a9a9e55b3

acharlieh:
@alecxvs it came from trying to prevent people from accidentally taking unreleased code when pointing at the github repo (prior to when we were using Supermarket internally to upload cookbooks to chef servers) while we tried incremental changes in process for contributing to cerner_splunk. If you're interested in the history, check out cerner/cerner_splunk#41.

gravesb:
Not a big deal, just personal preference to follow other open source cookbooks (and the vast majority of git projects in general) that uses master for the latest reviewed/approved code.  Per Charlie's comment, given the supermarket, I also don't see a need for a branch that always points to the latest release.

+1

alecxvs:
If there are no objections, I would prefer merging into master and getting rid of stable at the conclusion of this pull request.

alecxvs:
I think you're right about waiting, we can rearrange the branches after closing this.

jdmurphy:
+1 then if we are just going to wait and rearrange later.

alecxvs:
Squashed 💥

PR #2, opened 2016-07-11
Implement Service Resource

alecxvs:
### Implementation of the splunk_service resource, for managing splunk as a system service

Actions include starting, restarting, and stopping Splunk directly as well as enabling and disabling Splunk as a system service for running on startup.

Because of concerns with open file ulimits in our existing cerner_splunk cookbook, and the service being the best place to manage these limits, the service resource also handles setting ulimit for a Splunk installation.

I also moved the unit test directory so that all the tests could be run with just 'rspec' (also matches the structure that most cookbooks tend to use).


README.md
```diff
@@ -29,23 +31,65 @@ Installs Splunk or Universal Forwarder.
...
+*Note: Multiple installations of Splunk may technically be possible, but currently only one can be enabled as a system service.*
+
+**The following actions share the same properties:**
+
+##### Action *:start*
+Starts the Splunk daemon if not already running.
+If the ulimit is changed, invokes a restart of the daemon at the end of the run.
+
+##### Action *:restart*
+Starts the Splunk daemon, stopping it if it is already running.
```
jdmurphy: Minor, but this seems a weird way to phrase this. Might be better like: `Stops the Splunk daemon, if it is running, and starts it again.`

libraries/platform_helpers.rb
```diff
@@ -25,5 +25,19 @@ def package_names
         }
       }
     end
+
+    # def service_names
```
jdmurphy: Remove commented out code?

libraries/service_helpers.rb
```diff
@@ -0,0 +1,58 @@
+module CernerSplunk
+  # Mixin Helper methods for the Splunk Service resource
+  module ServiceHelpers
+    include ResourceHelpers
+    require 'win32/service' if RUBY_PLATFORM =~ /mswin|mingw|windows/
+
+    # Reports if the Windows service for the current package exists
+    # @return [Boolean]
+    def service_exists
```
jdmurphy: Minor, can we add a `?` to the end of the function name?

libraries/service_helpers.rb
```diff
@@ -0,0 +1,58 @@
...
+      ::Win32::Service.exists?(process_name)
+    end
+
+    def splunk_bin_path
+      Pathname.new(install_dir).join('bin')
+    end
+
+    # Returns the process id of the daemon is running
+    # @return [Integer]
```
jdmurphy: Should be `# @return [Integer, FalseClass] ` right?

libraries/service_helpers.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    # Reports the limits for the current package's running process
+    # @return [String] limits output
+    def service_limits
+      pid = service_pid
+      pid ? Pathname.new("/proc/#{pid}/limits").read : false
+    end
+
+    # Inserts ulimit command into the startup script for the current package's service
+    def try_write_initd_ulimit(ulimit)
```
jdmurphy: Function name could use some work. Maybe `insert_initd_ulimit`?

.kitchen.yml
```diff
@@ -22,12 +22,14 @@ suites:
   - name: install
     run_list:
       - recipe[cerner_splunk_ingredient_test::splunk]
-      - recipe[cerner_splunk_ingredient_test::universal_forwarder]
   - name: uninstall
     run_list:
-      - recipe[cerner_splunk_ingredient_test::splunk]
```
gravesb: Don't we still need to install something to test the uninstall recipe works properly?

libraries/service_helpers.rb
alecxvs: I had an issue stubbing this method with a question mark, but I'll give it another shot

.kitchen.yml
alecxvs: I need to look at this again... It isn't necessary because I updated the uninstall recipe to install before uninstall, but I think the main reason for doing that instead of keeping them separate was because the install recipe is set up so that it triggers a delayed restart (updating the ulimit) and that caused issues when uninstalling because it would then try to restart the splunk that didn't exist. I don't remember if I had a solution for that, or if I just decided "don't uninstall in the same recipe you're updating splunk".

libraries/service_helpers.rb
alecxvs: That was my initial idea, but it also updates the existing ulimit and if the file doesn't exist it doesn't actually change anything so I had a hard time coming up with a name that actually made sense for it.

spec/unit/helper_spec.rb
```diff
@@ -0,0 +1,164 @@
+require_relative 'spec_helper'
+include CernerSplunk::PathHelpers, CernerSplunk::RestartHelpers, CernerSplunk::ResourceHelpers
```
jdmurphy: This might just be me, but I would prefer separate test files for every helper to isolate everything.

libraries/service_helpers.rb
gravesb: Why not return nil or 0 rather than false?  It seems a bit weird to have a mixed return type of Integer or Boolean.

libraries/service_helpers.rb
gravesb: Same comment for line 33 (service_running?)

resources/splunk_install.rb
```diff
@@ -60,25 +50,41 @@ def package_path
...
   load_current_value do |desired|
     package desired.package
-    version_file = version_pathname(install_dir)
-    current_value_does_not_exist! unless version_file.exist?
 
-    version_data = version_file.read
-    installed_version, installed_build = version_data.match(/VERSION=(\d(?:\.\d)+).+BUILD=([\w]+)/m).captures
-    version installed_version
-    build installed_build
+    user_resource = Chef::Resource::User.new desired.user || new_resource.user, run_context
```
gravesb: minor, can we put some parens around the constructor args here?

.kitchen.yml
gravesb: Ah, I see.  Yeah, you just modified the test recipes.  That's fine, I just hadn't made it that far through the review yet.

spec/unit/helper_spec.rb
```diff
@@ -0,0 +1,164 @@
...
+      end
+    end
+  end
+end
+
+describe 'ResourceHelpers' do
+  describe 'load_installation_state' do
+    let(:name) { 'splunk' }
+    let(:install_dir) { '/opt/splunk' }
+    # class Node
```
jdmurphy: No longer needed?

spec/unit/helper_spec.rb
```diff
@@ -0,0 +1,164 @@
...
+    #     @run_state = {}
+    #   end
+    # end
+    let(:node) do
+      node = Chef::Node.new
+      node.automatic['os'] = 'linux'
+      node.automatic['kernel'] = { 'machine' => 'x86_64' }
+      node
+    end
+    let(:install_dir_path) { double('Pathname.new(install_dir)') }
```
jdmurphy: Is the double meant to have the construction of a pathname as a string in it?

spec/unit/install_examples.rb
```diff
@@ -1,12 +1,12 @@
 shared_examples 'should install' do |platform, expected_url|
   it 'should install' do
-    expect(run_state['current_installation']).to eq(
-      name: package.to_s,
-      package: package,
-      version: '6.3.4',
-      build: 'cae2458f4aef',
-      x64: true
-    )
+    # expect(run_state['current_installation']).to eq(
```
jdmurphy: No longer needed?

spec/unit/service_spec.rb
```diff
@@ -0,0 +1,135 @@
...
+
+                    allow_any_instance_of(Chef::Resource)
+                      .to receive(:shell_out)
+                      .with('sh -c "ulimit -n"', user: 'fauxhai')
+                      .and_return(user_limit_shout)
+                  end
+
+                  expect_any_instance_of(Chef::Provider).to receive(:clear_restart) if action == 'restart'
+
+                  # if !running && ['start', 'restart', 'enable'].include?(action)
```
jdmurphy: Remove commented code?

spec/unit/helper_spec.rb
alecxvs: Yeah, I named it to represent the object it's mocking. I could name it 'install_dir' if that makes more sense.

libraries/service_helpers.rb
alecxvs: I might have been using ||= at one point, but apparently not now. I should be able to return nil.
Returning zero would be easier, but I want something falsey (and not interpretable as a valid number/pid)

spec/unit/service_spec.rb
alecxvs: 0fafd8fe3c9e332ab5ac7d41dca2ba840999497e

spec/unit/helper_spec.rb
alecxvs: 0fafd8fe3c9e332ab5ac7d41dca2ba840999497e

spec/unit/helper_spec.rb
alecxvs: 0fafd8fe3c9e332ab5ac7d41dca2ba840999497e

libraries/service_helpers.rb
alecxvs: I can change to service_exists? but not service_running? and I can't figure out why:
```
Failure/Error:
       ChefSpec::SoloRunner.new({ step_into: ['splunk_service'] }.merge!(runner_params)) do |node|
         node.normal['test_parameters'] = test_params
         node.normal['run_state'].merge!(mock_run_state)
       end.converge('cerner_splunk_ingredient_test::service_unit_test')
     
     SyntaxError:
       splunk_service[splunk] (cerner_splunk_ingredient_test::service_unit_test line 5) had an error: SyntaxError: /Users/as032696/.rvm/rubies/ruby-2.1.8/lib/ruby/2.1.0/forwardable.rb:181: syntax error, unexpected '(', expecting ';' or '\n'
             def __service_running?_without_any_instance__(*args, &block)
                                                           ^
       /Users/as032696/.rvm/rubies/ruby-2.1.8/lib/ruby/2.1.0/forwardable.rb:181: syntax error, unexpected &
             def __service_running?_without_any_instance__(*args, &block)
                                                                   ^
```

alecxvs: 31c954aa1eaaa27c264d333d219f0eb51ad93f64
I didn't change service_running, maybe I can revisit that when I do test refactoring. That's a big maybe though because I suspect this error is occuring because I'm stubbing on Chef::Resource which is very early and it's trying to do some kind of Ruby magic to it and it's just not taking it (hence the question mark in the middle of the method), and whether or not I need to stub Resource or Provider is still a mystery to me, but I know that only one or the other works depending on where it gets called in rspec (first load or after other tests have run), or neither (chef run already happened).

README.md
alecxvs: 2e0d718db2b3b0ca50d262c8857872a74ae64bad

resources/splunk_install.rb
alecxvs: 2e0d718db2b3b0ca50d262c8857872a74ae64bad

libraries/service_helpers.rb
gravesb: I agree with Josh here. The primary goal of this method is to set the ulimit for the service, the fact that it doesn't do anything if the service isn't setup doesn't have to be reflected in the method name.  That's what the method comment is for.  Otherwise we would be putting try_ in front of a large number of methods we write.

resources/splunk_install.rb
```diff
@@ -60,25 +50,41 @@ def package_path
...
   load_current_value do |desired|
     package desired.package
-    version_file = version_pathname(install_dir)
-    current_value_does_not_exist! unless version_file.exist?
 
-    version_data = version_file.read
-    installed_version, installed_build = version_data.match(/VERSION=(\d(?:\.\d)+).+BUILD=([\w]+)/m).captures
-    version installed_version
-    build installed_build
+    user_resource = Chef::Resource::User.new(desired.user || new_resource.user, run_context)
```
gravesb: Doesn't this resource have to be added to the resource collection at some point? I'm missing the connection where this actually gets executed.

resources/splunk_service.rb
```diff
@@ -0,0 +1,158 @@
...
+# Resource for managing Splunk as a system service
+class SplunkService < ChefCompat::Resource
+  include CernerSplunk::PlatformHelpers, CernerSplunk::ResourceHelpers,
+          CernerSplunk::ServiceHelpers, CernerSplunk::RestartHelpers
+
+  resource_name :splunk_service
+
+  property :name, String, name_property: true, identity: true
+  property :package, [:splunk, :universal_forwarder], required: true
+  property :options, String, default: ''
```
acharlieh: what sort of arbitrary options would we allow here?

libraries/service_helpers.rb
alecxvs: 137d5d6f6a913d23a5b0355a57578e7c3d2d0cc7

libraries/service_helpers.rb
alecxvs: Documentation updated: 137d5d6f6a913d23a5b0355a57578e7c3d2d0cc7
Need to update method name

spec/unit/install_examples.rb
alecxvs: Removed: 137d5d6f6a913d23a5b0355a57578e7c3d2d0cc7

libraries/service_helpers.rb
```diff
@@ -14,34 +14,33 @@ def splunk_bin_path
...
         end
       end
-      false
     end
 
     # Reports if the daemon for the current package is running
     # @return [Boolean]
     def service_running
-      service_pid != false
+      !!service_pid # rubocop:disable Style/DoubleNegation
```
acharlieh: Don't just disable the style check: `!service_pid.nil?`

resources/splunk_service.rb
alecxvs: `--no-prompt` and `--answer-yes` are a couple I had to use at one point.

libraries/service_helpers.rb
alecxvs: Now that I have tests I can probably do that; I was trying to generally convert to a boolean representing truthiness, not just testing for nil. If false were returned from service_pid, it would return true here.

resources/splunk_service.rb
acharlieh: Both of those have to do with accepting the EULA and performing first run activities as opposed to launching the service. Should that be a different action for the service resource? Or does it fit with another resource? (Or at the first point we would execute a command, we can look for the ftr file in the installation directory and do it for them?)

libraries/service_helpers.rb
alecxvs: 135d6855e26a4540aaf39c35a30a653eebd70c05

libraries/service_helpers.rb
alecxvs: 135d6855e26a4540aaf39c35a30a653eebd70c05

libraries/platform_helpers.rb
alecxvs: 135d6855e26a4540aaf39c35a30a653eebd70c05

resources/splunk_service.rb
alecxvs: I'm handling `--accept-license` already. I didn't think those flags were first run only, if that's the case there's no point having options there. I might add no-prompt to my first run commands, and hopefully any issues that involve prompting at first run can be resolved outside of that prompt.

resources/splunk_service.rb
```diff
@@ -0,0 +1,158 @@
...
+                  shell_out('sh -c "ulimit -n"', user: user).stdout.to_i
+                end
+        ulimit limit
+      end
+    end
+  end
+
+  action :start do
+    ensure_license_agreement user
+    execute_cmd('start', user)
```
acharlieh: This feels a bit odd, we're enabling a system service with enable / disable, but then we're bypassing and executing splunk start, stop, and restart directly?

resources/splunk_service.rb
alecxvs: 184f73387dd43fb827f510eb58110b16e626bf6c

resources/splunk_service.rb
alecxvs: I did that to begin with because the Windows splunk service runs splunkd directly which doesn't seem ideal since running splunk directly does some sanity checks. I also thought that bin/splunk ran the init.d script if it was enabled, and that the service command didn't actually work for some reason, both of which I tested again and I've found them to be false.

I suppose then I should use the vanilla `service` resource to handle Splunk if it's enabled on linux systems.

alecxvs: Ahh, I'm not as crazy as I thought. So here's an interesting case. Splunk has to be started through the init script or else the service can't be stopped properly. Seems to be dependent on this `/var/lock/subsys/splunk` file existing.

The implementation of checking for this file doesn't reside in the init script itself, so that's kind of a bummer. I was hoping to add another condition somewhere to check for the pid file. I can probably remedy this by touching the lock file if I detect the service is running when enable is called.

alecxvs: 87114e5cea6b2df14999e0749beb98cda8bd9089

alecxvs:
Foodcritic is failing for using the service command...
```
FC004: Use a service resource to start and stop services: ./resources/splunk_service.rb:126
FC004: Use a service resource to start and stop services: ./resources/splunk_service.rb:136
FC004: Use a service resource to start and stop services: ./resources/splunk_service.rb:150
FC004: Use a service resource to start and stop services: ./resources/splunk_service.rb:163
```

But Chef is not smart enough to use the SysV scripts on Debian 8.4 and ends up directly terminating the Splunk process instead, causing bad state in the system service. I'm not exactly sure how I should resolve this, but I have a few ideas.

- I could see if the FC check can be disabled or otherwise trick it by using shell_out instead.
- If I were reasonably sure this was a Debian specific issue, I might be able to give a very specific provider to force it to handle it as SysV, but I'm not sure about either of those.
- I could ship my own Systemd unit file and install that over using enable/disable boot-start, but that could hide the enabled state from Splunk (which prevents running on a user other than the boot-start user when enabled, which seems important to have).

I'm not sure of Chef's logic behind using the last resort provider and hitting at the process directly. Hopefully I can figure out another way to make Chef realize that we have both SysV AND Systemd to work with here.

alecxvs:
After talking with @acharlieh on Friday, I think it would be easier to enable the system service on first run, and assume that we're going to use the system service from then on. It eliminates a lot of edge cases since I don't have to consider running Splunk without a service. This means I'm going to be using only the init.d script for setting ulimits and I'm ripping out the enable and disable actions. It's going to reduce a lot of code, so I'm going to squash my commits after I get this change committed.

alecxvs:
Refactoring complete, squashed commits.

libraries/service_helpers.rb
```diff
@@ -0,0 +1,58 @@
+module CernerSplunk
+  # Mixin Helper methods for the Splunk Service resource
+  module ServiceHelpers
+    include ResourceHelpers
+    # require 'win32/service' if RUBY_PLATFORM =~ /mswin|mingw|windows/
```
gravesb: delete commented code

.travis.yml
```diff
@@ -10,4 +10,4 @@ rvm:
 script:
 - bundle exec rubocop
 - bundle exec foodcritic -f any .
-- bundle exec rspec test/unit
```
gravesb: this needs to stay unless you move the specs from test/unit to the default spec directory.

gravesb:
I missed this in the last review, but if you specify RuntimeError in the raise_error matchers on line 71,96 and 119 of the install_spec it will suppress the console output of the errors. (e.g. `expect { chef_run }.to raise_error(RuntimeError, /Package must be specified.*/)`

libraries/service_helpers.rb
alecxvs: 3612f484be7f85fad25aaf21e49b18d1571ff1cc

alecxvs:
I think I tried doing that and it didn't suppress. I'll double check.

alecxvs:
Sure enough, the errors are suppressed. I must have done something differently.
cc253b27e8c1fc630ef2588ad1efb736d5efe8fe

.kitchen.yml
```diff
@@ -5,29 +5,34 @@ driver:
 provisioner:
   name: chef_zero
   require_chef_omnibus: 12.9
+  # log_level: debug
```
jdmurphy: Can we remove this?

alecxvs: 1cc5cb91c6dca436ce463072ff4cd9de1cab8032

spec/unit/resource_helpers_spec.rb
```diff
@@ -0,0 +1,87 @@
+require_relative 'spec_helper'
+include CernerSplunk::PathHelpers
```
gravesb: Shouldn't this include CernerSplunk::ResourceHelpers?  how is this working?

alecxvs: It must be working because the libraries are being mixed in on the top level and rspec is using the same object space across all the spec files, so because it's included in another spec file it usually works here. I say usually because if you only run this file with `rspec spec/unit/resource_helpers_spec.rb`, it will completely fail.

alecxvs: 645c9ef7265db3f64ab2c8c95d4d09e60a7bdfae

spec/unit/service_spec.rb
```diff
@@ -0,0 +1,140 @@
+require_relative 'spec_helper'
+include CernerSplunk::PathHelpers, CernerSplunk::PlatformHelpers, CernerSplunk::ServiceHelpers, CernerSplunk::RestartHelpers
```
gravesb: Why are these necessary?

gravesb:
Can we reorganize the specs a bit to something like:
```
spec/
  libraries/
    *_helpers_spec.rb
  resources/
    install_examples.rb
    splunk_install_spec.rb
    splunk_service_spec.rb
  spec_helper.rb
```
And yes, my suggestion also includes a rename for the resource specs to match the original file name.

spec/unit/resource_helpers_spec.rb
gravesb: Then it would probably be better if all these includes in the specs were moved into the initial describe block to keep from polluting the global object space. For example, the path_helpers_spec would look like:
```ruby
require 'spec_helper'

describe CernerSplunk::PathHelpers do
  include CernerSplunk::PathHelpers

  describe 'filename_from_url' do
    let(:expected_filename) { 'my_filename.tgz' }
    it 'should return the filename in a URL' do
      url = 'https://cool.website.internet/place/my_filename.tgz'
      expect(filename_from_url(url)).to eq(expected_filename)
    end
  end
end
```

spec/unit/service_spec.rb
alecxvs: Loading the libraries before Chefspec does so that it uses my stubs

alecxvs:
Rearranged the tests: a487960348f0a94254ac58402593acd053f88ed2

Unfortunately I had to keep them under spec/unit/ because foodcritic was reading spec/ and assumed things in there were supposed to be Chef DSL.

spec/unit/service_spec.rb
gravesb: Ok, a comment here would be helpful then.  especially if we move all the other global namespace includes into the specs. (I was able to get all of them working except for this one with the include in the initial describe block)

spec/unit/resources/service_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+  %w(start stop restart).each do |action|
+    describe "with action :#{action}" do
+      platform_package_matrix.select { |k, _| %w(windows redhat).include? k }.each do |platform, versions|
+        versions.each do |version, packages|
+          describe "on #{platform} #{version}" do
+            packages.each do |package, _|
+              context "with #{package}" do
+                let(:runner_params) { { platform: platform, version: version, user: 'root' } }
+                let(:test_params) { { name: package.to_s, action: action.to_sym } }
+                let(:install_dir) { default_install_dirs[package][is_windows ? :windows : :linux] }
```
gravesb: Are all of these let blocks needed?  I'm having a hard time finding where install_dir and cwd are used.

gravesb:
Fair enough, not sure how to tell foodcritic where to look for lint.  You could still rename the resource files to include the splunk_ prefix.

spec/unit/resource_helpers_spec.rb
alecxvs: I don't think it's possible. Running individual specs fails in any case after doing this. It doesn't seem to mix-in the code in the right place.

alecxvs: Oh I might be a bit off on that, it does work in some places... but I can't refer to the module itself, so the stubs no longer work. I don't know if there is another way to stub these methods.

alecxvs:
Yeah, forgot to do that. Made that change here: 65e9cbecdea5b687a7d57d0cf00e85e4fcfdd862

Running foodcritic: `foodcritic . -f any` from the project base directory

spec/unit/service_spec.rb
alecxvs: 65e9cbecdea5b687a7d57d0cf00e85e4fcfdd862

spec/unit/resource_helpers_spec.rb
alecxvs: 65e9cbecdea5b687a7d57d0cf00e85e4fcfdd862

spec/unit/resources/service_spec.rb
alecxvs: They don't seem to be used, must have accidentally abstracted them away.
c084647325989f218ef4fbf1d32ad95e81c1b59d

spec/unit/libraries/path_helpers_spec.rb
```diff
@@ -1,9 +1,11 @@
-require_relative 'spec_helper'
+require_relative '../spec_helper'
```
gravesb: minor, next time you're in here these can be changed to just `require 'spec_helper'`.  Then if they move around again it doesn't matter what directory they're in in relation to the spec helper file.

spec/unit/libraries/resource_helpers_spec.rb
```diff
@@ -0,0 +1,87 @@
+require_relative '../spec_helper'
+
+describe 'ResourceHelpers' do
+  include CernerSplunk::ResourceHelpers, CernerSplunk::PathHelpers
```
gravesb: minor, should only need ResourceHelpers here.  PathHelpers can be removed.

gravesb:
+1

jdmurphy:
+1 pending Brett's last comments being resolved.

spec/unit/libraries/path_helpers_spec.rb
alecxvs: That doesn't work
```
/Users/as032696/Git/cerner_splunk_ingredient/spec/unit/libraries/path_helpers_spec.rb:1:in `require': cannot load such file -- spec_helper (LoadError)
	from /Users/as032696/Git/cerner_splunk_ingredient/spec/unit/libraries/path_helpers_spec.rb:1:in `<top (required)>'
	from /Users/as032696/.rvm/gems/ruby-2.2.5/gems/rspec-core-3.5.1/lib/rspec/core/configuration.rb:1435:in `load'
	from /Users/as032696/.rvm/gems/ruby-2.2.5/gems/rspec-core-3.5.1/lib/rspec/core/configuration.rb:1435:in `block in load_spec_files'
	from /Users/as032696/.rvm/gems/ruby-2.2.5/gems/rspec-core-3.5.1/lib/rspec/core/configuration.rb:1433:in `each'
	from /Users/as032696/.rvm/gems/ruby-2.2.5/gems/rspec-core-3.5.1/lib/rspec/core/configuration.rb:1433:in `load_spec_files'
	from /Users/as032696/.rvm/gems/ruby-2.2.5/gems/rspec-core-3.5.1/lib/rspec/core/runner.rb:100:in `setup'
	from /Users/as032696/.rvm/gems/ruby-2.2.5/gems/rspec-core-3.5.1/lib/rspec/core/runner.rb:86:in `run'
	from /Users/as032696/.rvm/gems/ruby-2.2.5/gems/rspec-core-3.5.1/lib/rspec/core/runner.rb:71:in `run'
	from /Users/as032696/.rvm/gems/ruby-2.2.5/gems/rspec-core-3.5.1/lib/rspec/core/runner.rb:45:in `invoke'
	from /Users/as032696/.rvm/gems/ruby-2.2.5/gems/rspec-core-3.5.1/exe/rspec:4:in `<top (required)>'
	from /Users/as032696/.rvm/gems/ruby-2.2.5/bin/rspec:22:in `load'
	from /Users/as032696/.rvm/gems/ruby-2.2.5/bin/rspec:22:in `<main>'
	from /Users/as032696/.rvm/gems/ruby-2.2.5/bin/ruby_executable_hooks:15:in `eval'
	from /Users/as032696/.rvm/gems/ruby-2.2.5/bin/ruby_executable_hooks:15:in `<main>'
```

spec/unit/libraries/resource_helpers_spec.rb
alecxvs: 5e7252765dc4d85817b581206c51c60c8f54281b

spec/unit/libraries/path_helpers_spec.rb
gravesb: Ah right, I had it working when I moved all the specs into the root directory before I found out about the whole foodcritic thing and we put them all under /unit.  disregard my comment.

README.md
```diff
@@ -29,23 +30,54 @@ Installs Splunk or Universal Forwarder.
...
+|:--------|:-----------------------------------:|:--------:|:----------------------------------------------------------------|:-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| package | `:splunk` or `:universal_forwarder` | **Yes**  |                                                                 | Specifies the Splunk package to install. You must specify the package, or name the resource for the package; for example, `package :splunk` or `splunk_install 'universal_forwarder' do ... end` |
+| user    |               String                |    No    | Owner of the specified Splunk installation, or the current user | User to run Splunk as. This is the user that will be used to run the Splunk service.                                                                                                             |
```
acharlieh: Do we want people to specify the user here too? Or do we just want whatever is set in the install resource to win?

resources/splunk_install.rb
```diff
@@ -3,26 +3,22 @@
...
-  include CernerSplunk::PlatformHelpers, CernerSplunk::PathHelpers
+  include CernerSplunk::PlatformHelpers, CernerSplunk::PathHelpers, CernerSplunk::ResourceHelpers
 
   resource_name :splunk_install
 
   property :name, String, name_property: true, identity: true
   property :package, [:splunk, :universal_forwarder], required: true
   property :version, String, required: true
   property :build, String, required: true
+  property :user, String, default: lazy { node['current_user'] || package == :splunk ? 'splunk' : 'splunkforwarder' }
```
acharlieh: I think this should just default to 'splunk' 

node['current_user'] will always be 'root' on linux as it's the effective UID of the chef-client process: https://github.com/chef/ohai/commit/fec9467a0c93b203e97447207352745536c2b471

resources/splunk_install.rb
```diff
@@ -60,25 +50,43 @@ def package_path
...
-    current_value_does_not_exist! unless version_file.exist?
 
-    version_data = version_file.read
-    installed_version, installed_build = version_data.match(/VERSION=(\d(?:\.\d)+).+BUILD=([\w]+)/m).captures
-    version installed_version
-    build installed_build
+    user_resource = Chef::Resource::User.new(desired.user || new_resource.user, run_context)
+    user_resource.system true
+    user_resource.manage_home true
+    user_resource.run_action :create
```
acharlieh: Should we be creating a user in at the point where we're "loading the current values" ?

README.md
alecxvs: I'm not sure what you're asking about exactly. It will determine which user the install resource used by the owner of the splunk directory... except for on Windows, Chef can't seem to find that information. I'm not sure if we have a directory permissions issue on Windows, though. Besides that, I also think it's possible to run Splunk as a user that has permissions to the directory but is not the owner, so I don't necessarily want to restrict it to only the determined install user.

resources/splunk_install.rb
alecxvs: I thought it was possible to run chef-client as not root?
I suppose it wouldn't hurt anything to run forwarders as the splunk user, too.

resources/splunk_install.rb
alecxvs: It does seem wrong to do this here, but load_current_value is the only block of code called before every action. The alternative is to create a method in the action_class that I call from every action (I had to do something similar with post_install except after the action is run). I will probably do that, as this code shouldn't reside here, I just wish there was a better way.

alecxvs: Turns out I was calling super() on the beginning of every install action so it's easier than I thought.
888b6e98128e6320d82de74e2de2e9d9f3db9013

README.md
acharlieh: The knee-jerk reaction here was to use the user used by the install resource (probably communicated by the run_state)... Given how Splunk will write files to it's own directory, it can be problematic to run splunk as a different user than an installed user... For example if you do the first time run step as the root user, instead of the Splunk user, then start splunk as the splunk user, certain created files would not be accessible by Splunk and you wind up in a problematic state.

resources/splunk_service.rb
```diff
@@ -0,0 +1,109 @@
...
+    def service_action(desired_action)
+      service service_name do
+        provider Chef::Provider::Service::Systemd if systemd_is_init?
+        supports start: true, stop: true, restart: true, status: true
+        action desired_action
+      end
+    end
+
+    def configure_service
+      # Use run_action to execute immediately and avoid a race condition with /etc/init.d/splunk not existing yet.
```
acharlieh: Could you go into this? What's the race condition? Does it have to do with the setting of ulimit values?

alecxvs: Yeah, it's involved with the ulimit values.

Particularly in an action on the Linux provider, I have something like this:
```Ruby
configure_service

converge_if_changed :ulimit do
  write_initd_ulimit ulimit
end
```

configure_service should be calling enable boot-start and generating the init.d script in time for write_initd_ulimit (which writes the ulimit into the init.d script), but it does not. Calling run_action will execute the resource immediately, the same way you would call run_action to execute a resource at compile time instead of converge time.

I don't know why this happens exactly, but I suspect it has something to do with DSL evaluation, and converge_if_changed might play a part in that.

README.md
```diff
@@ -29,23 +30,54 @@ Installs Splunk or Universal Forwarder.
...
+| ulimit  |               Integer               |    No    | Start up script ulimit or user ulimit                           | Open file ulimit to give Splunk. This sets the ulimit in the start up script (if it exists) and for the given user in `/etc/security/limits.d/`. -1 translates to `'unlimited'`                  |
```
acharlieh: Thinking about [the most recent Virtual .conf](http://www.meetup.com/Splunk-Meetups/events/232707683/) I just watched, I wonder if we want this to me more than just the open file handle limit. But that probably could wait until later.

acharlieh:
+1 I think we'll have changes come up as we integrate but for now

README.md
alecxvs: Yeah, that would be something I would want to split out later.

alecxvs:
Squashed commits

PR #3, opened 2016-08-10
Implement Configuration Resource

alecxvs:
Implementation of the basic configuration resource.


alecxvs:
Need to update the README

alecxvs:
Also need to actually write the conf files as the specified user... whoops

libraries/conf_helpers.rb
```diff
@@ -0,0 +1,66 @@
...
+  # Mixin Helper methods for the Splunk Config resource
+  module ConfHelpers
+    def stringify_config(config)
+      config.map do |section, props|
+        [section.to_s, props.map { |k, v| [k.to_s, v.to_s] }.to_h]
+      end.to_h
+    end
+
+    def read_config(conf_path)
+      return {} unless Pathname.new(conf_path).exist?
```
gravesb: minor, could store a reference to the Pathname.new here and reuse it below on line 16.

alecxvs: c9ba089213de12f560044fb7a4d3fec854f7b63e

alecxvs:
Apparently I was already setting file owner properly, I just wasn't testing it.
I moved the file writing logic out of the helper so it was easier to test and ensure the ownership of.
45df017

Forgot to actually test the user in that commit, so I fixed some tests and found the user default value was not working properly (I knew there must have been a reason I wasn't using default). That should be fixed now.
183e2b1

README.md
```diff
@@ -78,6 +107,106 @@ Properties:
...
+
+### splunk_conf
+Manages configuration for an installation of Splunk. Requires splunk_install to be evaluated first.
+
+##### Action *:configure*
+Applies configuration to .conf files.
+You should know where the .conf file is that you wish to modify, as you must provide the path from `$SPLUNK_HOME/etc`.
+
+You can specify the scope as you would in the real path to the file (`system/local/indexes.conf`) or use the `scope`
+property and specfiy this path: `system/indexes.conf`. The resource will modify local config if no scope is provided.
```
gravesb: minor *specify

README.md
```diff
@@ -78,6 +107,106 @@ Properties:
...
+two = 2
+three = 3
+
+[stanza]
+key = value
+```
+
+##### Run State
+Since there is no concept of type in .conf files, everything is evaluated as strings. Configuration is read as strings
+and the provided config is converted to strings before merging. `splunk_conf` will store config (after merging) into the
```
gravesb: why do we need the config in run_state?

README.md
```diff
@@ -78,6 +107,106 @@ Properties:
...
+```Ruby
+node.run_state['splunk_ingredient'] = {
+  'current_installation' => {
+    name: 'splunk',
+    package: :splunk,
+    version: '5.5.5',
+    build: 'a1b2c3d4e5f6',
+    x64: true,
+    config: {
+      'testing' => {
```
gravesb: This example isn't accurate.  It should have the path as a parent key to the configs contained in that path.

README.md
alecxvs: It's another one of those things that we decided would be nice to have in the run state for consumers of the resource, though I don't know of any use case for it. It was trivial to implement, but I can just as easily remove it if we don't think we'll need it. One could instead use the read_config helper method to get the actual current configuration in a file.

README.md
alecxvs: 05cdf0a5c79950e4630c6db15428253e91c057e7

README.md
alecxvs: 05cdf0a5c79950e4630c6db15428253e91c057e7

README.md
alecxvs: Discussed offline, the run state is being loaded too late to be of any value, and consumers can load the config hash with the helper methods. The config run state has been removed.

9f402df59de69b8d8f9857e260d7ebfc64ce2e96

libraries/conf_helpers.rb
```diff
@@ -0,0 +1,62 @@
...
+          when /^\d*\.\d+$/
+            Float(value)
+          else
+            value
+          end
+        end
+      end
+    end
+
+    def merge_config(desired_config, current_config)
```
acharlieh: nit-picky -> flip the order of the arguments to match the .merge arguments

libraries/conf_helpers.rb
```diff
@@ -0,0 +1,62 @@
...
+      (current_config.keys + desired_config.keys).uniq.each do |section|
+        (current_config[section] ||= {}).merge! desired_config[section] if desired_config[section]
+      end
+
+      stream = StringIO.new
+      stream.puts '# Warning: This file is managed by Chef!'
+      stream.puts '# Comments will not be preserved and configuration may be overwritten.'
```
acharlieh: Nit-picky... usual style in other installs is to have the [default] stanza members just be at the top of the file. Do we want to do that here as well? (Thought being someone familiar with Splunk would be more familiar with that style)

libraries/conf_helpers.rb
alecxvs: 55adbbfdc640867c062950bc8388116ffc8cf34c

libraries/conf_helpers.rb
alecxvs: I would rather err on being more explicit to which section the properties belong to. I don't see a good reason to try and preserve style in a Chef managed file either as I'm already dropping all comments and whitespacing in favor of a consistent output. The logic to even preserve top-of-file config would be rather complex if I wanted it to merge properly with a default stanza. Even if I assumed all default config was destined for the top of the file, somebody who writes config to [default] would likely be unhappy about moving their config to a less concise location.

alecxvs:
Added integration tests, and discovered some issues with setting the user (that I corrected)
e561848

gravesb:
+1

jdmurphy:
+1

acharlieh:
+1

PR #4, opened 2016-09-02
Refactor Unit Tests

alecxvs:
Refactored the resource specs to use one-liners where possible, and use cached chef runs for all contexts. To aid in this, I opted to wrap the context method and provide a cached chef_run before the context is evaluated. This enables using one-liners (which generate their own examples) without a performance hit.

Unfortunately I could not refactor one-liners in everywhere, I realize now their only possible use is testing the state of a specific object. It does not work in any case where there are multiple or various subjects or where we are only testing that methods are called or errors are thrown. Thus, I did not touch the library specs. None of them have a case for one-liners or chef runs to be cached.

### DIFF
--- a/spec/unit/resources/splunk_conf_spec.rb
+++ b/spec/unit/resources/splunk_conf_spec.rb
@@ -3,7 +3,7 @@ include CernerSplunk::ConfHelpers, CernerSplunk::ResourceHelpers
 
 shared_examples 'splunk_conf' do |platform, version, package|
   let(:runner_params) { { platform: platform, version: version, user: 'root' } }
-  chef_context 'action :configure' do
+  chef_describe 'action :configure' do
     let(:config) { { a: { 'foo' => 'bar', 'one' => 1 } } }
     let(:existing_config) { { 'a' => { 'foo' => 'bar' } } }
     let(:expected_state_config) { { 'a' => { 'foo' => 'bar', 'one' => 1 } } }
@@ -209,8 +209,8 @@ describe 'splunk_conf' do
   let(:test_recipe) { 'config_unit_test' }
 
   environment_combinations.each do |platform, version, package, _|
-    describe "on #{platform} #{version}" do
-      describe "with package #{package}" do
+    context "on #{platform} #{version}" do
+      context "with package #{package}" do
         include_examples 'splunk_conf', platform, version, package
       end
     end
--- a/spec/unit/resources/splunk_conf_spec.rb
+++ b/spec/unit/resources/splunk_conf_spec.rb
@@ -1,6 +1,4 @@
-
 require_relative '../spec_helper'
-
 include CernerSplunk::ConfHelpers, CernerSplunk::ResourceHelpers
 
 shared_examples 'splunk_conf' do |platform, version, package|
@@ -207,8 +205,6 @@ shared_examples 'splunk_conf' do |platform, version, package|
 end
 
 describe 'splunk_conf' do
-  include CernerSplunk::PathHelpers
-
   let(:test_resource) { 'splunk_conf' }
   let(:test_recipe) { 'config_unit_test' }
 
--- a/spec/unit/resources/splunk_conf_spec.rb
+++ b/spec/unit/resources/splunk_conf_spec.rb
@@ -4,8 +4,8 @@ require_relative '../spec_helper'
 include CernerSplunk::ConfHelpers, CernerSplunk::ResourceHelpers
 
 shared_examples 'splunk_conf' do |platform, version, package|
-  describe 'action :configure' do
-    let(:runner_params) { { platform: platform, version: version, user: 'root' } }
+  let(:runner_params) { { platform: platform, version: version, user: 'root' } }
+  chef_context 'action :configure' do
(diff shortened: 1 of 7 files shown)


spec/unit/resources/splunk_conf_spec.rb
```diff
@@ -4,8 +4,8 @@
 include CernerSplunk::ConfHelpers, CernerSplunk::ResourceHelpers
 
 shared_examples 'splunk_conf' do |platform, version, package|
-  describe 'action :configure' do
-    let(:runner_params) { { platform: platform, version: version, user: 'root' } }
+  let(:runner_params) { { platform: platform, version: version, user: 'root' } }
+  chef_context 'action :configure' do
```
jdmurphy: I feel like contextually this doesn't work. In rspec, although `describe` and `context` are functionally the same, contextually they are differentiated in that `describe` tests against one functionality while the `context` is testing functionality against a given state.

I know you wrote the chef_context to get around the weirdness around contexts, but we might need a similar chef_describe if we want to be contextually clear.

spec/unit/resources/splunk_install_spec.rb
```diff
@@ -4,121 +4,121 @@
...
   include CernerSplunk::PathHelpers, CernerSplunk::PlatformHelpers
 
-  let(:runner_params) { { platform: 'redhat', version: '7.1' } }
-  let(:test_params) { { name: 'splunk', build: 'cae2458f4aef', version: '6.3.4' } }
+  let(:test_resource) { 'splunk_install' }
+  let(:test_recipe) { 'install_unit_test' }
 
-  let(:windows_opts) { 'LAUNCHSPLUNK=0 INSTALL_SHORTCUT=0 AGREETOLICENSE=Yes' }
+  environment_combinations.each do |platform, version, package, expected_url|
+    describe "on #{platform} #{version}" do
```
jdmurphy: Similar to my [comment above](https://github.com/cerner/cerner_splunk_ingredient/pull/4/files#r77647371), I think these seem like contexts and not describes since this is describing state.

spec/unit/resources/splunk_service_spec.rb
```diff
@@ -1,137 +1,146 @@
 require_relative '../spec_helper'
 
-# TODO: Change how we're stubbing methods during refactoring so we can move this out of the global object space.
 include CernerSplunk::ServiceHelpers, CernerSplunk::RestartHelpers
```
gravesb: Any update here?  One of the goals of the refactor was to not have to include these in the global space.

alecxvs: Moving stubs to the ChefSpec::SoloRunner block did not alleviate this.

acharlieh:
Out of curiosity, part of the goal of this was to increase the performance (speed of execution) of the unit tests? It's interesting then that [latest stable build](https://travis-ci.org/cerner/cerner_splunk_ingredient/builds/153057623) seems to have taken 18 minutes versus [the latest build of this PR](https://travis-ci.org/cerner/cerner_splunk_ingredient/builds/157209720) took 29 minutes. I'm +1 whatever's decided, but it's an odd data point and seems to go counter to what chef-run caching was supposed to solve.

alecxvs:
That's a really good point, it should have the same performance if not better. I see it spitting out a bunch of warnings about using compat_resource, I'll try updating gems and seeing if i can replicate that

spec/unit/resources/splunk_conf_spec.rb
alecxvs: 952a76fd23eb7d3cc8b19ec62d1d463c2bda6ab6

I would've aliased the method to make it simpler, but I'm using context in the method implementation and contexts can't be used top level.

spec/unit/resources/splunk_install_spec.rb
alecxvs: 952a76fd23eb7d3cc8b19ec62d1d463c2bda6ab6

spec/unit/resources/splunk_service_spec.rb
alecxvs: Some follow up to this,
I'm not totally concrete on this but I think stubbing things under Chef::Provider has issues, where Chef::Resource is fine with mixins in the describe block. That explains why it works in splunk_install_spec and not conf or service.

I'm actually surprised it works any differently at all, if it doesn't just outright not work. The block should be called at the top level due to the nature of blocks, but when mixing into the block, the libraries don't show up anywhere in the object space in RSpec. I haven't found any discussion or explanation of mixins in a block, probably because few people have been daring enough to try it.

alecxvs:
I did some performance tests with let vs. cached, locally it makes quite a difference (50 sec vs. 35 sec cached), but travis.ci seems very inconsistent. The tests seem to take longer in general, with non-cached running in 24 minutes, but the second cached run (to make sure it wasn't due to some kind of availability) ran in 35 minutes, compared to the 27 minutes of the first run. I'm chalking this up to travis.ci running on mostly shared resources and being inconsistent as a result.

gravesb:
+1

nmattam:
+1

jdmurphy:
+1

PR #5, opened 2016-09-15
Refactor "ensured restart" to a resource

alecxvs:
Refactored the ensured restart methods into a resource for consumption by a recipe.

I left the helper methods in for splunk_service, because it's very difficult to use/test in load_current_value (no DSL support, instantiated resource classes don't exist in the resource collection thus are invisible to Chefspec). I took the simplest approach, also maintaining passivity with the existing unit tests, and instantiated the resources in the helper methods.

I've yet to run integration tests.


spec/unit/resources/splunk_restart_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+          it { is_expected.to create_file_if_missing(Pathname.new(install_dir).join('restart_on_chef_client').to_s) }
+
+          it 'should notify the Splunk service to restart' do
+            expect(subject.splunk_restart(package.to_s)).to notify("splunk_service[#{package}]").to(:restart).delayed
+          end
+        end
+
+        chef_describe 'action :check' do
+          let(:test_params) { { name: package.to_s, action: :check } }
+          let(:action_stubs) do
```
gravesb: minor, let's wrap lines 37-45 in a chef_context 'when the marker exists'

alecxvs: f2dac9b7483c803dba9363a80527c0293a9fa3d2

nmattam:
+1

jdmurphy:
+1

Gemfile
```diff
@@ -1,17 +1,18 @@
 source 'https://rubygems.org'
 
 gem 'berkshelf', '~> 4.3'
-gem 'chef', '~> 12.4'
+gem 'chef', '~> 12.4', '<= 12.13.37'
```
alecxvs: Is there another reason to lock this down? My reason was that newer versions deprecate compat_resource, so it spits a bunch of warnings, but that's not a very solid reason as it seems to still test okay.

nmattam: I ran into another issue with cerner_splunk where it pulled in 12.14.60 version of chef gem which required ruby>2.2.0 which broke the travis build.

alecxvs: Alright then, I'll add that reason. My issue is fixed in https://github.com/chef-cookbooks/compat_resource/pull/108

alecxvs: I made a change in 986d14d to support Chef 12.4 by testing on Ruby 2.2

.travis.yml
```diff
@@ -6,8 +6,16 @@ cache: bundler
 sudo: false
 bundler_args: --without local
 rvm:
-- 2.1.8
+  - 2.1.8
+  - 2.2.5
```
gravesb: we should update this to 2.3.1, that's what chef 12.14 packages as the embedded ruby.

alecxvs: 417dc5357d255e3146b32dad1854b8e3a9a6f282

gravesb:
+1 again after the travis changes to include multiple rubies and chef versions.

PR #7, opened 2016-09-30
Add Proc evaluation to config

alecxvs:
Added proc/lambda evaluation to the config resource and helpers. It can evaluate a proc at the top-level (handles the entire config hash) or at the stanza-level (handles the stanza and its contents).


test/cookbooks/cerner_splunk_ingredient_test/recipes/configure_splunk.rb
```diff
@@ -1,15 +0,0 @@
-splunk_conf 'system/indexes.conf' do
```
alecxvs: This file apparently wasn't being referenced anywhere. I presume it's leftover from when I was splitting the resources into their own recipes, but I forgot to clean it up previously.

spec/unit/libraries/conf_helpers_spec.rb
```diff
@@ -4,6 +4,65 @@
...
+        'second' => 'true'
+      },
+      'other' => {
+        'something' => 'here'
+      }
+    }
+  end
+
+  describe 'evaluate_config' do
+    context 'with top-level proc' do
```
jdmurphy: We might as well  use `let(:subject) { evaluate_config(existing_config, config) }` since we are testing this multiple times.

jdmurphy:
+1

gravesb:
+1

nmattam:
+1

PR #8, opened 2016-10-17
More Enhancements

alecxvs:
- Documentation is added explaining that splunk_restart invokes splunk_service of the same name
- Added tests for setting the name property of splunk_restart (turns out we technically already had a property to do this)
- Documentation is added for the group property on install and conf resources.
- ConfHelpers is made class-level, and thus available to recipes
- New method in ConfHelpers to parse configuration directly from a string (or IO objects, anything with an each_line method)


test/cookbooks/cerner_splunk_ingredient_test/recipes/conf_helpers.rb
```diff
@@ -0,0 +1,24 @@
+my_config = {
```
gravesb: Given that these are just ruby methods, I don't think we need to cover them with test kitchen directly. Just straight up rspec tests should suffice.

test/cookbooks/cerner_splunk_ingredient_test/recipes/restart_unit_test.rb
```diff
@@ -1,11 +1,14 @@
...
-splunk_service params['name'] do
+resource_name = params.delete('resource_name')
+
+splunk_service params['name'] || resource_name do
+  package params['package'] if params['package']
   action :nothing
 end
 
-splunk_restart params['name'] do
+splunk_restart resource_name do
```
gravesb: minor, shouldn't this be the same as above? `params['name'] || resource_name`

test/cookbooks/cerner_splunk_ingredient_test/recipes/conf_helpers.rb
alecxvs: I want to test that they can be called from a recipe, where they previously could not be.

alecxvs:
After talking with Brett offline, we've determined that there is no use case for ConfHelper in external cookbooks. The only uses in cerner_splunk are in the splunk_app resource which will be replaced. While I would like to expose these methods to consumers, the only feasible way I could see that happening is if we packaged ConfHelpers in a gem.

I removed the Chef tests for ConfHelpers, but I kept it purely singleton. It doesn't need to be a mix-in because it is Chef agnostic (other helpers depend on Chef or resource state) and this greatly reduces complexity in unit testing and usage where these helpers are involved. For consistency, I've also made PathHelpers purely singleton since it is also Chef agnostic.

libraries/conf_helpers.rb
```diff
@@ -48,5 +52,5 @@ def merge_config(current_config, desired_config)
       end
       stream.string
     end
-  end
+  end unless defined?(ConfHelpers)
```
gravesb: This seems a bit odd.  What's this here for?

alecxvs: Chef still does weirdness to the objectspace by reloading these files at converge time. This is necessary for stubbing in Chefspec by ensuring the module is only defined once.

spec/unit/resources/splunk_conf_spec.rb
```diff
@@ -1,5 +1,6 @@
 require_relative '../spec_helper'
-include CernerSplunk::ConfHelpers, CernerSplunk::ResourceHelpers
+# require_relative '../../../libraries/conf_helpers'
```
gravesb: minor, remove commented code

gravesb:
+1

spec/unit/libraries/conf_helpers_spec.rb
```diff
@@ -108,26 +106,44 @@
         }
       }
     end
-    let(:conf_file) { double('Pathname.new(conf_path)') }
 
-    it 'should return the config in a hash' do
-      expect(Pathname).to receive(:new).with(conf_path).and_return(conf_file)
-      expect(conf_file).to receive(:exist?).and_return(true)
-      expect(conf_file).to receive(:readlines).and_return(IO.readlines('spec/reference/read_test.conf'))
+    it 'should call parse_config' do
```
jdmurphy: I am not sure this test name is very descriptive of what we are actually testing since I don't see anything about parse in the test itself.

jdmurphy:
+1

spec/unit/libraries/conf_helpers_spec.rb
alecxvs: You're right. There used to be a stub here but now it makes no sense.
acef0eb7718669c6b363f62f20b832769e5aeb80

PR #9, opened 2016-11-02
Add install_dir property and archive installation

alecxvs:
Added install_dir property to all resources and enabled forcing installation from an archive on Windows and Linux.

Integration tests on Windows are currently failing, Splunk is loading some Python which is throwing random memory errors. I think I need to expand the memory size of the Windows VMs.


alecxvs:
Integration tests are now passing

spec/unit/resources/splunk_install_spec.rb
```diff
@@ -130,5 +146,80 @@
...
+
+              chef_context 'when install_dir is provided' do
+                let(:install_dir) { platform == 'windows' ? 'C:\\Splunk' : '/etc/splunk' }
+                let(:test_params) { { resource_name: package.to_s, build: 'cae2458f4aef', version: '6.3.4', install_dir: install_dir } }
+
+                it { is_expected.to unpack_poise_archive(package_path).with(destination: install_dir) }
+              end
+            end
+          end
+          chef_describe 'action :uninstall' do
```
gravesb: minor, add a newline above this describe block

test/integration/archive_install/spec.rb
```diff
@@ -0,0 +1,48 @@
+windows = os.windows?
+
+control 'archive_install' do
```
gravesb: Minor, these control blocks are optional and are more intended for when the inspec test is covering an official compliance check of some sort.  Since these are just custom tests I don't think we need them.

alecxvs: d62826838b8342190f6907ecb0839e1e5968cc3b

gravesb:
+1

nmattam:
+1

jdmurphy:
+1

.kitchen.yml
```diff
@@ -16,7 +16,8 @@ platforms:
   - name: debian-8.4
   - name: windows
     driver:
-      box: opentable/win-2012r2-standard-amd64-nocm
+      box: mwrock/Windows2012r2
+      gui: false # This box apparently defaults to using a GUI.
```
alecxvs: He reassures me that running it without gui is fine, and he often runs it this way as well. It's just a flag to determine whether or not virtualbox opens a gui, so I assume in his workflows he often wants a gui to open from running vagrant vms.

acharlieh (approved):
+1

.kitchen.yml
mwrock: yeah its just the most commonly desired setting so I put it their to make that easy.

acharlieh: Thanks @mwrock! I'm likely just an odd bear then, tending to prefer headless. But if the masses have spoken. 👍 

(nit-picky comment for @alecxvs then: The comment on this line should be clarified to reflect this fact maybe something like `# Most users of this box want the GUI, therefore it defaults to having one that we don't need.` Knowing myself, I am liable to comment on it again in a future review forgetting that we know the reasoning behind the box setting and we're not just going "well it appears this way" )

PR #10, opened 2016-12-08
Refactor converge_if_changed usage

alecxvs:
Prevent race conditions by not running resources in the converge_if_changed block.

Added a new helper class for the provider level (to be included in action_class). This is necessary to use Chef DSL objects like converge_if_changed. Instead of relying on the converge_if_changed block to execute resources when version or build numbers are changed, I've implemented a method that returns true or false based on the execution of converge_if_changed. Most of the previously non-DSL resources are now being instantiated into the Chef resource collection instead of only existing as pure Ruby. These cases should also now be unit tested where they previously could not be.

### DIFF
--- a/resources/splunk_install.rb
+++ b/resources/splunk_install.rb
@@ -88,8 +88,7 @@ class SplunkInstall < ChefCompat::Resource
     def post_install
       ruby_block 'load_version_state' do
         block { load_version_state }
-        only_if { changed? }
-      end
+      end if changed?
 
       execute "chown -R #{user}:#{group} #{install_dir}" do
         not_if { node['os'] == 'windows' || current_owner == user }
@@ -155,8 +154,7 @@ class ArchiveInstall < SplunkInstall
 
     poise_archive package_path do
       destination install_dir
-      only_if { changed? :version, :build }
-    end
+    end if changed? :version, :build
 
     post_install
   end
@@ -181,8 +179,7 @@ class RedhatInstall < SplunkInstall
     rpm_package package_name do
       source package_path.to_s
       action :install
-      only_if { changed? :version, :build }
-    end
+    end if changed? :version, :build
 
     post_install
   end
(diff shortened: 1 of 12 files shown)


resources/splunk_install.rb
```diff
@@ -146,10 +153,9 @@ def kernel_string
   action :install do
     super()
 
-    converge_if_changed :version, :build do
-      poise_archive package_path do
-        destination install_dir
-      end
+    poise_archive package_path do
+      destination install_dir
+      only_if { changed? :version, :build }
```
gravesb: this is functionally identical right?  was this part changed just for consistency with other sections?

alecxvs: Not exactly. I changed this for the same reason, being that resources inside converge_if_changed tend to cause race conditions. If I remember correctly, in this case the :delete notification on the above remote_file would fire without waiting for the archive extraction to finish.

test/cookbooks/cerner_splunk_ingredient_test/recipes/universal_forwarder.rb
```diff
@@ -7,7 +7,7 @@
 splunk_conf 'system/server.conf' do
   config(
     general: {
-      serverName: 'test-forwarder'
+      invalid: 'test-forwarder'
```
alecxvs: The actual hostname. I've discovered it's not being overwritten, it's not setting the config in the first place. I think doing `only_if { changed? }` in these sub resources is evaluating changed properties for those sub resources.

alecxvs: 95295ed5f484621891429a7fab7b9b7bbc95d531

nmattam:
+1

jdmurphy:
+1

gravesb:
+1

PR #11, opened 2017-01-19
Initial implementation for splunk_app

alecxvs:
- Base resource for splunk_app (no runtime provider swapping yet)
- splunk_app_custom implementation for Chef defined apps
- Some reformatting of the README using Remark
- Added value-level evaluation for config
- Added install path to the run state of existing installations (splunk_conf would previously assume a default directory based on the current installation's package)


resources/splunk_app.rb
```diff
@@ -0,0 +1,130 @@
...
+  def app_path
+    @app_path ||= Pathname.new(splunk_app_path).join(name)
+  end
+
+  def config_scope
+    'local'
+  end
+
+  def parse_meta_access(access)
+    access_read = Array(access[:read] || access['read']).join(', ')
```
nmattam: Are we expecting that every stanza in the *.meta file will have _access_ defined? I'm wondering what will happen if we only have something like 

```
[views]
export = system
```

resources/splunk_conf.rb
```diff
@@ -10,7 +10,7 @@ class SplunkConf < ChefCompat::Resource
   property :path, [String, Pathname], name_property: true, desired_state: false, identity: true
   property :install_dir, String, required: true, desired_state: false
   property :package, [:splunk, :universal_forwarder], required: true, desired_state: false
-  property :scope, [:local, :default], desired_state: false, default: :local
+  property :scope, [:local, :default, :none], desired_state: false, default: :local
```
nmattam: Should we update the README?

README.md
```diff
@@ -287,11 +307,79 @@ Removes the file marker.
...
+| configs      |                 Proc                |              No             |                                             | Proc containing `splunk_conf` resources. The resources will behave similarly to normal, except the config will be placed relative to the app directory rather than relative to the Splunk installation, and you can not change the scope. |
+| files        |                 Proc                |              No             |                                             | Proc containing miscellaneous resources for manipulating the app directory. This is ideally directories, templates, and files. The Proc will be given an absolute app path parameter for you to use in these resources.                  |
+| metadata     |                 Hash                |              No             | {}                                          | Configuration to apply to the metadata (local.meta in most cases, default.meta for custom apps). See splunk\_conf for more info, and also below for special handling of the `access` key.                                                |
```
alecxvs: It applies to local.meta in every case except for custom apps.

resources/splunk_app.rb
alecxvs: I suspect it will throw a method not found error. I also noticed that it's not even checking if access is a hash before trying to manipulate it.

README.md
acharlieh: If it's an app we're creating on the fly, then we would edit default.meta (and /default configs), if it's a packaged app or app we're downloading from another source, then local is what we use: ( to quote @duckfez parodying Arthur C Clarke - "all these worlds are yours, except /default - attempt no editing there" )

libraries/resource_helpers.rb
```diff
@@ -17,6 +17,10 @@ def splunk_bin_path
       Pathname.new(install_dir).join('bin')
     end
 
+    def splunk_app_path
```
acharlieh: If we're only calling this within the single app resource, why are we breaking it out into a helper? 

(also this is for locally installed apps as opposed to SHC Deployer, Indexer Cluster Master, or Deployment Server pushed apps).

resources/splunk_app.rb
alecxvs: 79edba309310d3c36f5a685e6975b53d762b014b

resources/splunk_conf.rb
alecxvs: 79edba309310d3c36f5a685e6975b53d762b014b

libraries/resource_helpers.rb
alecxvs: I liked the idea of putting it next to the similar splunk_bin_path... but you're right, it's only going to be used in splunk_app, and it's probably going to disappear when clustering support is added.

libraries/conf_helpers.rb
```diff
@@ -2,9 +2,13 @@ module CernerSplunk
   # Helper methods for reading and evaluating Splunk config
   module ConfHelpers
     def self.evaluate_config(current_config, desired_config)
-      return desired_config.call(current_config) if desired_config.is_a? Proc
+      desired_config = desired_config.call(current_config) if desired_config.is_a? Proc
       desired_config.map do |section, props|
-        props.is_a?(Proc) ? props.call(section, current_config[section] || {}) : [section, props]
+        mapout = props.is_a?(Proc) ? props.call(section, current_config[section] || {}) : [section, props]
+        mapout[1] = mapout[1].map do |key, value|
+          value.is_a?(Proc) ? value.call(key, (current_config[section] || {})[key]) : [key, value]
```
acharlieh: Do we want to talk about the proc arguments? why push current_config as an argument? Also should each layer include the keys from above? Should the top be app and stanza name? What are the set of procs you forsee at each and why would they use these keys?

Also at each layer, do we want to bug out if it's a String instead of a map?

libraries/resource_helpers.rb
alecxvs: a5c70ea4e58559de13da73ae3058baa417340a87

libraries/conf_helpers.rb
alecxvs: I will log a JIRA to enhance the config to provide robust objects for reading config in Procs.
Another concern is that there is no way to delete from the config, only 'blank' it which won't remove stanzas or keys. That's something to note on the new JIRA as well, it's a bit out of scope here.

spec/unit/resources/splunk_app_spec.rb
```diff
@@ -0,0 +1,425 @@
...
+      'splunk_ingredient' => {
+        'installations' => {
+          install_dir => install
+        },
+        'current_installation' => install
+      }
+    }
+  end
+
+  # %w(splunk_app_custom splunk_app_package splunk_app_git).each do |resource, test|
```
gravesb: delete commented code

alecxvs:
Added Integration Tests and fixed some bugs.

libraries/conf_helpers.rb
alecxvs: JIRA for these changes: https://jira2.cerner.com/browse/HIINFOPS-1250

spec/unit/resources/splunk_app_spec.rb
```diff
@@ -0,0 +1,425 @@
...
+
+            include_examples 'app_install'
+
+            chef_context 'when version is the same' do
+              let(:version_config) { { 'launcher' => { 'version' => '2.0.0' } } }
+
+              include_examples 'app_no_install'
+            end
+          end
+          chef_context 'when version is not provided' do
```
gravesb: minor, newline before the context block

gravesb:
+1

resources/splunk_app.rb
```diff
@@ -0,0 +1,142 @@
...
+      raise 'Version to install must be specified when app has a version.' unless desired.version
+      version app_version
+    end
+  end
+
+  action_class do
+    include CernerSplunk::ProviderHelpers
+
+    def apply_config
+      node.run_state['splunk_ingredient']['conf_override'] = {
```
acharlieh: should this be in a ruby block as well?

resources/splunk_app.rb
```diff
@@ -0,0 +1,142 @@
...
+        conf_path: Pathname.new('apps').join(name).join(config_scope).to_s,
```
acharlieh: Should this leverage splunk_app_path instead of being hard coded to apps ?

resources/splunk_app.rb
```diff
@@ -0,0 +1,142 @@
...
+
+class CustomApp < SplunkApp
+  resource_name :splunk_app_custom
+
+  def config_scope
+    'default'
+  end
+
+  action :install do
+    return unless !version || changed?(:version)
```
acharlieh: Wouldn't we want to do these actions always with a custom app? not only if version changed?

alecxvs: This is only applicable when the app has a version, which our custom apps will likely not have.

resources/splunk_app.rb
alecxvs: It would certainly be much more predictable that way.

34ac1da5b99cfd8fbef13c440abaa43f69eb28be

I suspect I could switch it up and do something in after_created instead so that blocks aren't necessary, but this works for now.

resources/splunk_app.rb
alecxvs: That's a bit tricky... It should not be hard coded to 'apps', but splunk_app_path does not work here (being an absolute path, also including '/etc/' so relative_path_from(install_dir) does not work). This can be changed along with 'apps' in splunk_app_path when cluster support is implemented.

spec/unit/resources/splunk_app_spec.rb
alecxvs: a2a9b6df729ea4e411b184ed6707c422bffd20ca

alecxvs:
Nibin spotted that I was checking local/app.conf for the version instead of default/app.conf
7d5a6366ab4ad435a3f3a24b03c839def93b7d7a

nmattam:
I'm seeing this while converging app-install-centos-67 https://gist.github.com/nmattam/d43959d1f2e048433f0cfa83b6a5c951

alecxvs:
Whoops. Looks like I can't use ruby_blocks after all...

alecxvs:
I changed up splunk_conf overrides to be read at resource compile time, so removed the ruby blocks. Also dealt with a bad regex string (unsure how this passed the first time).

There's another problem I realized; the owner of the splunk installation (directories and files) is not being set for resources in the files proc (I don't have magic control of these) and there's no decent way to provide this to consumers currently. I set the owner, knowing that it is 'splunk' by default for the splunk package, but hopefully I can find some way to set this in the background in the future.

test/cookbooks/cerner_splunk_ingredient_test/recipes/splunk_app.rb
```diff
@@ -0,0 +1,27 @@
...
+      config(
+        debug: {
+          banana: 'yellow'
+        }
+      )
+    end
+  end)
+
+  files(proc do |app_path|
+    file Pathname.new(app_path).join('plain_file.txt').to_s do
```
nmattam: Can create a file like [nav.xml](https://github.com/cerner/cerner_splunk/blob/stable/vagrant_repo/data_bags/cerner_splunk/cluster-apps-vagrant.json#L66-L67), which resides few directories deep? I'm guessing we cannot because the parent directories (data/ui/views/) are not being created here.

Also if I'm understanding this, _files_ property can be used only to create files directly under the app directory. Are there files that we create under the app directory? Shouldn't they all be folders under the app directory (I might be wrong though)?

nmattam:
+1 pending https://github.com/cerner/cerner_splunk_ingredient/pull/11/files#r98234737

test/cookbooks/cerner_splunk_ingredient_test/recipes/splunk_app.rb
alecxvs: The files property calls a proc and gives the app path as a parameter, and that's it. Consumers are intended to use any number of file and directory resources inside this proc to create non-config files. It doesn't make any assumptions about the contents or try to enforce it, so you could create the data/ui/views directory in the same proc and then create that file.

acharlieh:
+1 for an initial pass.

As I review, it feels odd to have to explicitly call proc, instead of just interpolating the block as a proc. 

e.g. the difference between:
```
splunk app 'foo' do
  files(proc do
    ...
  end)
end
```

and 
```
splunk app 'foo' do
  files do
    ...
  end
end
```

But as I also look at it, and I think about the issues of forcing owner / group / perms I wonder if files and configs are in fact just one in the same, and we just follow the pattern of setting useful configuration on the run state with the surrounding ruby blocks that you have ... like file permissions should be X, directory permissions should be X', you should use Y context, absolute path to the root of the app directory is Z...

These feel like improvements over the basic functionality though so +1 for now.

alecxvs:
I believe it could be implemented to call a block, but those stop being resource properties and must be methods instead. But it is feasible.

I wonder if it would be possible to have some kind of localized patching... where, when you call file or directory or some other expected resource there, it runs through and sets up the resource as normal but before finishing it would modify the desired path, owner, and group to be relevant to the splunk installation.

Like you said though, those are definitely more improvements and we haven't finished the total implementation yet. This is something I can consider with the upcoming changes.

PR #12, opened 2017-02-16
Add splunk_app_package implementation

alecxvs:
- Introduced splunk_app_package, a resource capable of downloading and extracting tar.gz packages of Splunk apps.
- Removed user as a property of splunk_service. Since the user running Splunk should be the same as the user that owns Splunk, it will pull from the current owner user of the install directory.
- Windows did not properly support users, so it is more explicit in code that Windows does not support specifying users and will always use the current user of the Chef run (likely SYSTEM).
- Updated Rubocop to check against Ruby 2.3 syntax. This means adding `# frozen_string_literal: true` to every file. It made no difference functionally as there are no string literals being mutated in this project.


resources/splunk_app.rb
```diff
@@ -69,25 +65,43 @@ def parse_meta_access(access)
...
+
+      case strategy
+      when :prep
+        app_cache_path.mkpath
+        FileUtils.mv(app_path, app_cache_path + name)
+      when :keep_existing
+        FileUtils.cp_r(app_cache_path + name + 'local', app_path + 'local')
+        CernerSplunk::FileHelpers.deep_change_ownership(app_path + 'local', current_owner, current_group)
+
+        FileUtils.cp(app_cache_path + name + 'metadata/local.meta', app_path + 'metadata/local.meta')
```
nmattam: Shouldn't we do the same with [lookups](https://github.com/cerner/cerner_splunk/blob/stable/libraries/splunk_app.rb#L266-L268)?

alecxvs: We decided to prefer the apps lookups in any case... I don't remember if there was any case to preserve unique lookups, though.

nmattam:
Some unit tests are failing which is causing travis to fail.

resources/splunk_app.rb
```diff
@@ -136,3 +148,41 @@ def config_scope
...
+  def package_path
+    return @package_path if @package_path
+    file_cache = Pathname.new(Chef::Config['file_cache_path'])
+    @package_path = file_cache.join(CernerSplunk::PathHelpers.filename_from_url(source_url).gsub(/.spl$/, '.tgz'))
+  end
+
+  action :install do
+    return unless !version || changed?(:version)
+
+    remote_file package_path.to_s do
```
nmattam: There is a [check](https://github.com/cerner/cerner_splunk/blob/stable/libraries/splunk_app.rb#L199) for snapshot version installing over released version before downloading the file in the current splunk_app implementation in cerner_splunk. Also, we do some [validation](https://github.com/cerner/cerner_splunk/blob/stable/libraries/splunk_app.rb#L223-L239) over the downloaded tarball before installing it. Should we be including those checks?

alecxvs: Perhaps checking that the archived app is the same name as the app resource is a good check, but I'm not sold on the other cases. How would we unexpectedly get a prerelease version given a specific source url? And why do we care if the app has local directories or not?

@acharlieh

nmattam: > How would we unexpectedly get a prerelease version given a specific source url?

I will leave that to @acharlieh.

>  And why do we care if the app has local directories or not?

Everything that ships with the app should reside in the _default_ folder. _local_ is where the KOs created/modified by the consumers of the app will live. 
https://answers.splunk.com/answers/37729/deployment-config-files-in-local-or-default.html

nmattam:
+1 once we decide on how to go with the [validation](https://github.com/cerner/cerner_splunk_ingredient/pull/12#discussion_r102244778).

resources/splunk_app.rb
alecxvs: I guess what I was asking is, why does it matter that we fail for local config in an app when the consumer is specifically trying to use that app.

Properly handling that with the future upgrade strategy would complicate things though. It would be better to fail on that case.

acharlieh: > How would we unexpectedly get a prerelease version given a specific source url?

Easy... upload something to the wrong spot, or have an out of date working copy and build/upload. There wasn't and still really isn't much in the way of an ecosystem for building / packaging apps sadly, so I wasn't assuming any protections from an app build system in this

> Properly handling that with the future upgrade strategy would complicate things though. It would be better to fail on that case.

This is the core of why we have the fail for local content. I would support a flag saying that "no we expect there to be local content in this package" that is default set to fail instead of pass but yeah.

alecxvs: So, because I don't have a mechanism currently to inspect the tarball, I will need to extract it to a temporary directory before installing the packaged app. Because of this I can delete local content instead of failing outright. Do you think this would be preferable @acharlieh ?

alecxvs: Also, should the validation instead be on that the package version matches the desired version (since that is a parameter)? I could see it being a strict match to the desired version to make sure the downloaded app is exactly what is expected, or simply checking that the downloaded app is a release version if a release version is given. The version validation, I feel, are between the package's version and the version property, rather than against the installed version.

acharlieh: @alecxvs Extracting to a temp directory before installing is probably good (assuming you don't just lift the Tarball class from cerner_splunk).

But with the handling of local, I think the default behavior should be to fail on local content being present, with a parameter that if set could say "Yes, delete local content" or "No, keep local content" 

We should indeed be validating that the package version matches the expected version as _one of_ the checks. In cerner_splunk this is handled: https://github.com/cerner/cerner_splunk/blob/2.9.0/libraries/splunk_app.rb#L231-L233 
If the expected version does not match the version from the tarball exactly, then we fail unless we specified to expect a release version and the tarball contains a prerelease of the specified release version.

(We later handle the case that the release version is installed and we're trying to install a prerelease on top). Basically all of the guard conditions to make sure we don't have pain later

acharlieh: Per our offline discussion on BSD/Posix files... playing with the Tarball class locally I don't think cerner_splunk handles those either so I logged: cerner/cerner_splunk#143

resources/splunk_app.rb
```diff
@@ -69,25 +65,43 @@ def parse_meta_access(access)
     end
   end
 
+  # Provider exclusive methods
+
   action_class do
     include CernerSplunk::ProviderHelpers
 
+    def upgrade_app(strategy)
+      require 'fileutils'
```
acharlieh: I feel like for readability, any requires that are just built into ruby and will happen on all platforms we should have outside. Is there a reason to delay evaluation of this require?

resources/splunk_app.rb
```diff
@@ -33,11 +31,11 @@ def app_path
...
 
   def config_scope
-    'local'
+    @config_scope ||= (resource_name == :splunk_app_custom ? 'default' : 'local')
   end
 
-  def splunk_app_path
-    Pathname.new(install_dir).join('etc/apps')
+  def app_cache_path
+    @app_cache_path ||= Pathname.new(Chef::Config['file_cache_path']).join('splunk_ingredient/old_apps')
```
acharlieh: Should we be consistent on our style? Here we use `.join` but below we use `+`

resources/splunk_app.rb
```diff
@@ -69,25 +65,43 @@ def parse_meta_access(access)
     end
   end
 
+  # Provider exclusive methods
+
   action_class do
     include CernerSplunk::ProviderHelpers
 
+    def upgrade_app(strategy)
```
acharlieh: The name of this method is kinda confusing because it's not really performing an upgrade with a particular strategy if that make sense. What about splitting it into `backup_app`  and `restore_app_parts(strategy)` ? 

Should we also have a `remove_backup` ? as it seems we could wind up with a backup already present and then what happens when we move the app to that place?

resources/splunk_app.rb
acharlieh: We want to prefer the apps lookups but that's not a universal truth we need to give the choice on it. But also restoring the local data without restoring the lookup files, there could be transforms stanzas and permissions that would refer to a lookup that was built through the UI, that would stop functioning with this

libraries/file_helpers.rb
```diff
@@ -0,0 +1,20 @@
...
+      FileUtils.chown(desired_owner, desired_group, path.to_s)
+    end
+
+    def self.deep_change_ownership(path, owner, group = nil)
+      return if Chef.node.platform_family?('windows')
+
+      change_ownership(path, owner, group)
+      Pathname.glob(Pathname.new(path).join('**/*')).each { |sub_path| change_ownership(sub_path, owner, group) }
+    end
+  end unless defined? FileHelpers
```
acharlieh: I go back and forth on this unless pattern when it comes to defining modules that are the base versions of the Module... Should we do this? Or would we accidentally break if someone wanted to monkeypatch in a method and got sourced before us.

libraries/matchers.rb
```diff
@@ -43,4 +44,28 @@ def check_splunk_restart(name)
...
+
+  def install_splunk_app_custom(name)
+    ChefSpec::Matchers::ResourceMatcher.new(:splunk_app_custom, :install, name)
+  end
+
+  def install_splunk_app_package(name)
+    ChefSpec::Matchers::ResourceMatcher.new(:splunk_app_package, :install, name)
+  end
+
+  def install_splunk_app_git(name)
```
acharlieh: Do we have a git app already?

resources/splunk_app.rb
alecxvs: It probably makes no difference if this is required outside. This method is not called in every instance of the resource, however.

resources/splunk_app.rb
alecxvs: I'm not clear on how to handle the version validation. Ensuring that the tarball version matches the expected version should eliminate any concern of an unexpected version coming through the URL. Are you suggesting ignoring this check if the installed version is a prerelease version, and also having the validation to prevent switching to a prerelease version from a release version?

libraries/file_helpers.rb
alecxvs: It's necessary to use with Chefspec, and I'm not keen on changing this to allow monkeypatching.

resources/splunk_app.rb
alecxvs: We need to investigate our usage of apps to determine how we should approach this problem... if it's even a problem.

resources/splunk_app.rb
alecxvs: After offline discussion, I've learned that we want to validate that, if the desired version and installed version share the same release version, the desired version must not be a prerelease version.

This will require some parsing, and likely encapsulation in a new object.

libraries/matchers.rb
alecxvs: No, I can remove this.

resources/splunk_app.rb
alecxvs: I'm not sure there's an established style regarding pathnames, but I feel that situations including `Pathname.new(...)` or ending with method calls should continue to use methods such as `.join()`, and concatenation (`+`) is to be used with simple references. It is rather inconsistent in this file, and I will remedy this, but I'm not certain this is a specific case that should be changed.

resources/splunk_app.rb
alecxvs: I have some desire to contain these methods under one umbrella, for easy identification most of all. I think I can make these methods such as `backup` and `keep_existing` under a module `AppUpgrade`. Thoughts?

resources/splunk_app.rb
alecxvs: e17bdc8
5073a5a

libraries/matchers.rb
alecxvs: 620401feb83d4838959f8ec4e5c9882cd801cd58

resources/splunk_app.rb
jdmurphy: I guess I am unsure of how you are differentiating "simple" references vs others as it seems to me that consistency should be a goal.

jdmurphy:
+1

libraries/provider_helpers.rb
```diff
@@ -1,9 +1,88 @@
...
+        pkg_version = CernerSplunk::SplunkVersion.from_string(pkg_app_conf['launcher']['version'])
+
+        # Check that the package's version matches the desired base version.
+        unless pkg_version == app_version || !app_version.prerelease? && pkg_version.release_version == app_version.release_version
+          raise "Downloaded app version does not match intended version to install (#{pkg_version} vs. #{version})"
+        end
+
+        # Check that the package's version is not a pre-release when we really expect a release
+        if !app_version.prerelease? && pkg_version.prerelease?
+          raise "Downloaded app version was unexpectedly a pre-release version (#{pkg_version} vs. #{app_Version})"
```
alecxvs: 1d1d887738add97bd5b4c87300ecfa659f99a48d

.gitignore
```diff
@@ -6,6 +6,8 @@
 *.un~
 pkg/
 
+*.spl
```
alecxvs: 1d1d887738add97bd5b4c87300ecfa659f99a48d

resources/splunk_app.rb
alecxvs: Simple references would be something like `app_path + 'local'`
The case where I want to be locally consistent are like `Pathname.new('/opt/splunk').join('etc')` as opposed to `Pathname.new('/opt/splunk/') + 'etc'`
And cases where methods are being chained from the pathname object would almost require using join `app_path.join('local').mkpath` (I suppose one could do `(app_path + 'local').mkpath` but I'm not aware if thats stylistically acceptable)

There are some varying cases throughout the entire project... I would like to clean them up, but I don't entirely feel this is the time to do that. I will make an effort to make pathname joins consistent in the next PR.

I would like to continue this conversation as well, as the consensus might be to prefer using `+` or `.join` everywhere.

nmattam:
+1

resources/_splunk_app_upgrade.rb
```diff
@@ -0,0 +1,129 @@
...
+  property :name, String, name_property: true, identity: true
+  property :install_dir, String, required: true
+  property :version, [String, CernerSplunk::SplunkVersion], required: true
+
+  resource_name :_splunk_app_upgrade
+
+  default_action :nothing
+
+  def after_created
+    version CernerSplunk::SplunkVersion.from_string(version) if version
```
acharlieh: So an odd thing here... Property version is documented of taking a String or a SplunkVersion... so if I as a consumer pass a SplunkVersion, this would actually fail...

resources/splunk_app.rb
```diff
@@ -1,24 +1,26 @@
...
-  property :version, String
+  property :version, [String, CernerSplunk::SplunkVersion]
   property :configs, Proc
   property :files, Proc
   property :metadata, Hash, default: {}
 
   default_action :install
 
+  def after_created
+    version CernerSplunk::SplunkVersion.from_string(version) if version
```
acharlieh: Likewise... if version is a CernerSplunk::SplunkVersion ?

acharlieh (approved):

resources/_splunk_app_upgrade.rb
alecxvs: It will absolutely fail 😟

alecxvs: This file shouldn't actually exist... Oops!

alecxvs: fcb5bc680320d62b6f113c5f40da6f42a45052cb

resources/splunk_app.rb
alecxvs: fcb5bc680320d62b6f113c5f40da6f42a45052cb

PR #13, opened 2017-03-15
Add support for cluster apps

alecxvs:
Added a property to specify the root install directory for the app. This should typically be `:shcluster`, `:master_apps`, or `'deployment-apps'`. The first two are symbols as they are special cases. They have an apps directory (the string equivalent would be providing `'shcluster/apps'` or `'master-apps/apps'`) where deployment-apps does not.

I've also bundled some formatting changes to the README and style changes where pathname joins are being used. If you have any input on the pathname join style, please comment.


metadata.rb
```diff
@@ -4,7 +4,7 @@
 maintainer_email 'splunk@cerner.com'
 license          'Apache 2.0'
 description      'Installs and configures Splunk'
-long_description IO.read(File.join(File.dirname(__FILE__), 'README.md'))
+long_description((Pathname.new(__FILE__).dirname + 'README.md').read)
```
nmattam: Is a space missing here?

alecxvs: No, this is technically a method call and rubocop demanded method parentheses be used here.

nmattam:
+1

bbrownz:
+1

resources/splunk_app.rb
```diff
@@ -32,8 +33,18 @@ def install_state
...
   end
 
+  def absolute_app_root
+    root_path = case app_root
+                when :master_apps then 'master-apps/apps'
+                when :shcluster then 'shcluster/apps'
+                else app_root
+                end
+
+    Pathname.new(install_dir) + 'etc' + root_path
```
acharlieh: It might be worth a comment, that + is join, and that + absolute path means resolve the absolute path because this threw me for a brief loop.

acharlieh (approved):

resources/splunk_app.rb
alecxvs: 8908c06fd50066cca7a1cdf64301a289ba1dbe8a

PR #14, opened 2017-03-27
Enhance splunk_conf Proc API

alecxvs:
Added a context object to the proc parameters within splunk_conf.
Also, Rubocop wanted magic spaces for magic comments. The diff is now full of hot air. Better to do it now with a simple PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,12 +241,11 @@ end
 
 ###### Section-level Evaluation
 
-section-level evaluation allows you to evaluate the config key values for a specific section without being
+Section-level evaluation allows you to evaluate the config key values for a specific section without being
 responsible for regurgitating all of the config you didn't want to modify.
 
 In this example, only a specific section's contents are turned into lowercase strings. The rest is
-merged as usual. **Note that the section-level proc receives the context and a props hash, and expects
-a pair (array) of the desired section name and props hash to be returned.**
+merged as usual.
 
 ```Ruby
 splunk_conf 'system/test.conf' do
@@ -255,7 +254,7 @@ splunk_conf 'system/test.conf' do
       one: 1,
       four: 4
     },
-    section: ->(context, props) { [context.section.to_s.downcase, props.map { |k, v| [k.to_s.downcase, v.to_s.downcase] }.to_h] }
+    section: ->(context, props) { props.map { |k, v| [k.to_s.downcase, v.to_s.downcase] }.to_h }
   )
 end
 ```
@@ -267,14 +266,13 @@ ease of use. This is especially handy when you want to produce some value for a 
 nesting the rest of the section in your Proc. If there is a case where you're applying a Proc to values whose keys you don't know,
 this becomes a necessary functionality.
 
-In this example, if the value is an Array object, it is converted into a comma spaced string using join. **Note that the Value-level proc
-receives the context and value, and expects a pair (array) of the desired key and value to be returned (similar to the section-level proc).**
+In this example, if the value is an Array object, it is converted into a comma spaced string using join.
 
 ```Ruby
 splunk_conf 'system/test.conf' do
   config(
     testing: {
-      servers: ->(context, value) { [context.key, value.is_a?(Array) ? value.join(', ') : value] }
+      servers: ->(context, value) { value.is_a?(Array) ? value.join(', ') : value }
     }
   )
 end
@@ -282,7 +280,7 @@ end
 
 ###### The Context object
 
-A ConfContext object is passed as the first parameter to any splunk_conf proc. It contains contextual information such as
+A ConfContext object is passed as the first parameter to any splunk\_conf proc. It contains contextual information such as
 path of the current file, current section (for section-level and key-level procs), and current key (for key-level procs).
 It also attempts to determine the current app from the path, if any.
 
@@ -294,7 +292,7 @@ Consider the value-level example above:
 splunk_conf 'system/test.conf' do
   config(
     testing: {
-      servers: ->(context, value) { [context.key, value.is_a?(Array) ? value.join(', ') : value] }
+      servers: ->(context, value) { value.is_a?(Array) ? value.join(', ') : value }
     }
   )
 end
@@ -311,7 +309,6 @@ lambda do |context, value|
   [context.key, value]
 end
 ```
-
 
 ### splunk\_restart
 
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ These resources can:
 
 ## Requirements
 
-Chef >= 12.13.37
+Chef >= 12.14
 Ruby >= 2.3.1
 
 Supports Linux and Windows based systems with package support for Debian, Redhat,
@@ -30,7 +30,7 @@ Note: There are more specific resources available if you wish to override how Sp
 For example, if you want to force install from .rpm you can use splunk\_install\_redhat, or to install
 from archive (tgz for Linux and zip for Windows) use splunk\_install\_archive.
 
-#### Action *:install*
+#### Action _:install_
 
 Installs Splunk or Universal Forwarder.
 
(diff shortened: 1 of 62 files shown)


libraries/conf_helpers.rb
```diff
@@ -57,5 +66,28 @@ def self.merge_config(current_config, desired_config)
...
+      def initialize(path, stanza = nil, key = nil)
+        self.path = path
+        self.stanza = stanza
+        self.key = key
+      end
+
+      def path=(path)
+        @path = path
+        pathname = Pathname.new @path
+        @app = /local|default|metadata$/.match(pathname.parent.to_s) && pathname.parent.parent.basename.to_s
```
gravesb: As is, the $ only applies to metadata. Currently it's equivalent to `/(local)|(default)|(metadata$)/` but I think we want `/(local|default|metadata)$/`. need to add some parens to group it properly.

gravesb (changes requested):
There should be some readme updates under the advanced usage section since the procs now take different params.

libraries/conf_helpers.rb
```diff
@@ -57,5 +66,28 @@ def self.merge_config(current_config, desired_config)
...
+      end
+
+      def ==(other)
+        other.path == @path && other.app == @app && other.stanza == @stanza && other.key == @key
```
acharlieh: comparison of app is now unnecessary because if path does not match, then app will not match. if path matches then app matches.

libraries/conf_helpers.rb
acharlieh: Would the structure be more straightforward as: 
`@app = pathname.parent.parent.basename.to_s if regex.match(pathname.parent.to_s)`

Possibly premature regex optimization Should we also anchor the regex to a leading slash, and use a non-capturing group? e.g.  `[\/\\](?:local|default|metadata)$`

alecxvs: Good points. I was thinking of using a non-capturing group myself. I'm not sure I need to use a leading slash since I'm anchored to the end of the path.

libraries/conf_helpers.rb
alecxvs: dc1004343349c84dc0c1fee56005de86f4df01a6

libraries/conf_helpers.rb
alecxvs: dc1004343349c84dc0c1fee56005de86f4df01a6

alecxvs:
@acharlieh If I recall, we wanted to also make it so that nil values were removed from the config?

gravesb (approved):

gravesb:
+1.  Though I am a bit curious why the name change from stanza to section?

alecxvs:
The naming was inconsistent; I was referring to it as both stanza and section throughout... and section seems to be the generally correct way to refer to a section of an ini file. The Splunk docs themselves seem to be inconsistent, though they use stanza more than section. I could probably go either way, I just went with the more general name.

README.md
```diff
@@ -238,14 +239,14 @@ splunk_conf 'system/test.conf' do
...
-Stanza-level evaluation allows you to evaluate the config key values for a specific stanza without being
+section-level evaluation allows you to evaluate the config key values for a specific section without being
 responsible for regurgitating all of the config you didn't want to modify.
 
-In this example, only a specific stanza's contents are turned into lowercase strings. The rest is
-merged as usual. **Note that the Stanza-level proc receives the section string and a props hash, and expects
-a pair (array) of the desired section string and props hash.**
+In this example, only a specific section's contents are turned into lowercase strings. The rest is
+merged as usual. **Note that the section-level proc receives the context and a props hash, and expects
+a pair (array) of the desired section name and props hash to be returned.**
```
acharlieh: Out of curiosity, what's the usecase for letting the user change the section name dynamically with a Section level stanza? (And how does it work if I have a hash with one section with the name that it targets dynamically already set?)

acharlieh:
@alecxvs Yes... the thought being a top level proc generating a nil should result in the file being deleted, section level proc generating nil -> section deleted, value level proc generating nil -> value deleted. Any of these procs wanting an empty value should return an empty hash, or empty string respectively.

libraries/conf_helpers.rb
```diff
@@ -27,15 +36,15 @@ def self.read_config(conf_path)
...
-      current_stanza = 'default'
+      current_section = 'default'
       current_config = {}
 
       conf_body.each_line do |ln|
         case ln
         when /^\s*\[([^#]+)\]\s*/
-          current_stanza = Regexp.last_match[1]
+          current_section = Regexp.last_match[1]
         when /^\s*([^#]+?)\s*=\s*([^#]+?)\s*(?:#.*)?$/
```
acharlieh: We probably need a more robust conf reader at some point... as I have heard that others are using line continuation characters in configs

README.md
alecxvs: Yes, you're right. It made sense at the time, when stanza name and contents were being provided, to expect those to be returned. Since we have the context now, and every level of available evaluation, we don't need that.

libraries/conf_helpers.rb
alecxvs: 😱

README.md
alecxvs: Changed the expected return to be just the relevant config
2feb0485c7e0ec9ae8a45c0bf69323badbb7fd4f

alecxvs:
Added functionality to config_merge so if the config hash has a defined section or key with a nil value, it is deleted from the final config.

2feb0485c7e0ec9ae8a45c0bf69323badbb7fd4f

libraries/conf_helpers.rb
```diff
@@ -1,16 +1,28 @@
...
-        mapout
-      end.to_h
+    private_class_method def self.evaluate(config, context, data)
+      return config.call(context, data) if config.is_a?(Proc)
+      config
+    end
+
+    def self.evaluate_config(path, current_config, config)
+      context = ConfContext.new(path)
+      evaluted_config = evaluate(config, context.freeze, current_config) || {}
```
acharlieh: Nit spelling: evaluated_config

libraries/conf_helpers.rb
```diff
@@ -27,35 +39,74 @@ def self.read_config(conf_path)
...
         when /^\s*([^#]+?)\s*=\s*([^#]+?)\s*(?:#.*)?$/
-          (current_config[current_stanza] ||= {})[Regexp.last_match[1]] = Regexp.last_match[2]
+          (current_config[current_section] ||= {})[Regexp.last_match[1]] = Regexp.last_match[2]
         end
       end
 
       current_config
     end
 
     def self.merge_config(current_config, desired_config)
```
acharlieh: It took me a bit to understand where the filtering happens, but I found it here, and as I think through it, should this method be broken up? It seems like this one method does 3 different functions that might be better reallocated elsewhere?

1) Load and pull forward data from disk (which might be able to be one of many levels of procs)
2) Filter nulls (Which probably should be part of evaluate_config so decisions can be made based on the resultant map? )
3) Render the map to text to be written to disk (which would this be better as an ERB template? )

acharlieh:
I'm +1 other than the evaluation on if we should break up functionality a bit.

libraries/conf_helpers.rb
alecxvs: 1. merge_config isn't pulling data from disk, all config is passed in the two hash parameters.
2. Moving nil filtering to evaluate makes less sense. It does handle both current and desired config, but it only passes current config to the procs. merge_config deals with combining the current and desired configs and seems to be the most appropriate place for this. I could extract it into a method, but I don't see any value in that.
3. Using a template is a good idea... I can move the logic of building the conf file to the resource.

alecxvs: 10f9c8ad238395f5f34f8d298e122ec93ce26cda

I split the nil filtering into a new method, filter_config, and call it from the conf resource. Also the file writing logic has been moved to a template in the conf resource and merge_config now returns hash.

acharlieh (approved):

libraries/conf_helpers.rb
alecxvs: af47d72a035d0779e05786ed19acff9c6700607d

PR #18, opened 2017-05-02
Move splunk_restart functionality into splunk_service

### LINKED ISSUE #17: Restart marker file not clearing on start/stop
Calling `:start` or `:stop` on splunk_service does not clear the marker file laid by splunk_restart. The service may restart when a start (or redundant restart) is not desired. Additionally, we aren't checking for the presence of the restart marker file early enough. Running splunk_service inside another resource could have unexpected results.

alecxvs:
Fixes #17 

## Summary
These changes fold the functionality of splunk_restart into splunk_service. This aims to solve a problem with not clearing the restart marker on start and stop actions, and also to simplify our code.

The minimum Chef version was bumped from 12.14 to 12.16 due to the use of the newer resource method `#delayed_action`.

## Functional changes
Previously, the method to ensure that a delayed restart survived a failed chef run was to call the action :ensure on splunk_restart. With these changes, it is instead the :desired_restart action on splunk_service.

## Testing changes
- Tested that the marker file is being deleted on :start, :stop, and :restart actions.
  - Ensured that the marker file was **not** being deleted when the :start action notified for a :desired_restart upon noticing ulimit changes.
- Tested that :__guarded_restart is being notified from every created splunk_service resource.
- Tested that :desired_restart was placing the marker file.
- Tested that :__guarded_restart was running :restart and subsequently removing the marker file
- Changed the default parameters of splunk_service test resource to use a non-package name and explicit package parameter.
  - Added a test for package in the resource name to cover the old case.


resources/splunk_service.rb
```diff
@@ -22,12 +21,18 @@ class SplunkService < Chef::Resource
 
   def after_created
     package_from_name unless property_is_set?(:package) || property_is_set?(:install_dir)
+    delayed_action :__guarded_restart if run_context.parent_run_context.nil?
```
gravesb: What case is this meant to cover?  A comment with some explanation of why we're checking the parent run context would help here.

libraries/matchers.rb
```diff
@@ -32,20 +35,12 @@ def restart_splunk_service(name)
...
-  end
-
-  def ensure_splunk_restart(name)
-    ChefSpec::Matchers::ResourceMatcher.new(:splunk_restart, :ensure, name)
-  end
-
-  def check_splunk_restart(name)
-    ChefSpec::Matchers::ResourceMatcher.new(:splunk_restart, :check, name)
+  def desired_restart_splunk_service(name)
+    ChefSpec::Matchers::ResourceMatcher.new(:splunk_service, :desired_restart, name)
```
acharlieh: Nit: what's the ordering of these methods? i s d _ ?

resources/splunk_service.rb
```diff
@@ -62,14 +65,30 @@ def install_state
...
+    # Provide file and service resources from methods to eliminate accidental resource cloning
+    def marker_resource
+      resources(file: marker_path.to_s)
+    rescue Chef::Exceptions::ResourceNotFound
+      file(marker_path.to_s) do
+        action :nothing
+      end
+    end
+
+    def service_resource
```
acharlieh: Why do we have this indirection? Would it be more straightfoward to just create the resources in each action to do the right things/notifications? (Shouldn't they be a forked run so not worried about cloning? Or is that not true?)

resources/splunk_service.rb
```diff
@@ -62,14 +65,30 @@ def install_state
...
         provider Chef::Provider::Service::Systemd if systemd_is_init?
         supports start: true, stop: true, restart: true, status: true
-        action desired_action
+        action :nothing
       end
     end
 
+    def service_action(desired_action)
+      service_resource.run_action(desired_action)
+      marker_resource.run_action(:delete)
```
acharlieh: This relationship between the service resource and the marker resource is not correct, it's a notifies relationship, not an ordered run relationship. 

Let's say: I call service :start ... there are 2 possibilities here... either, I successfully transitioned to a start state, at which point i should remove the marker (in response to a notification of a state change)... or I was already started, no transition happened, thus no removal of the marker should take place.

acharlieh (changes requested):

resources/splunk_service.rb
alecxvs: You're right... I came to this implementation trying to fix resource cloning but I forgot that requirement.

resources/splunk_service.rb
alecxvs: In my testing, it was causing resource cloning due to multiple actions being run on the same service resource. I will investigate naturally created resources again in order to utilize notifications

resources/splunk_service.rb
alecxvs: 2e30d0f9001ba3d19bd40c1f36b147c065fde7f0

resources/splunk_service.rb
```diff
@@ -22,12 +21,18 @@ class SplunkService < Chef::Resource
...
+    delayed_action :__guarded_restart if run_context.parent_run_context.nil?
   end
 
   def service_name
     @service_name ||= service_names[package][node['os'].to_sym]
   end
 
+  def marker_path
+    @marker_path ||= Pathname.new(install_dir) + 'restart_on_chef_client'
+    # @marker_path
```
alecxvs: 2e30d0f9001ba3d19bd40c1f36b147c065fde7f0

libraries/matchers.rb
alecxvs: 2e30d0f9001ba3d19bd40c1f36b147c065fde7f0

resources/splunk_service.rb
alecxvs: 907b10c20b678dc615a1ea9ef90bb270d6e4f59c

resources/splunk_service.rb
alecxvs: I think you're suggesting declaring the service and file resources in every action they're used, which would equate to repeating those exact declarations with different actions 5 times (the linux specific implementation requires them to be redefined because the parent actions are not reused). In addition, I don't think it's actually doable. I tried declaring the resources normally in service_action (how it used to be, for the most part) but since I've started notifying actions on the resource's self, it is causing those to be cloned resources and throw warnings. The methods are defined in action_class so they should be scoped to the provider and declare those resources in the forked run context, but I don't know why each action run does not have its own run context. Regardless, the outcome is that all the actions being run on the service resource as part of checking and restarting off the restart marker are referencing the same run context so I wrote these methods to protect against resource cloning.

gravesb (approved):

resources/splunk_service.rb
```diff
@@ -114,10 +131,10 @@ class LinuxService < SplunkService
 
     if changed? :ulimit
       write_initd_ulimit ulimit
-      ensure_restart if service_running
+      run_context.notifies_immediately(Notification.new(current_resource, :desired_restart, current_resource)) if service_running
     end
 
-    service_action :start
+    service_action :start unless changed?(:ulimit) && service_running
```
acharlieh: I'm not sure if you need the guards here and on line 134... but there might be an interesting ordering thing between the immediate notification and the guards that you may want to check out.

acharlieh (approved):

PR #19, opened 2017-08-11
Fixes for cerner_splunk integration

alecxvs:
# What? Why?
Various tweaks to correct bugs and facilitate the integration into cerner_splunk

# How was it tested?
RSpec, Kitchen tests, Vagrant integration testing


.kitchen.yml
```diff
@@ -8,11 +8,11 @@ _script: &PACKAGE_APP_SCRIPT
           Darwin)
             echo '## Mac detected, using gtar'
             command -v gtar >/dev/null 2>&1 || { echo "You need to install gnu-tar to test app_upgrade. 'brew install gnu-tar'" >&2; exit 1; }
-            gtar -czvf test/files/pkg-app.spl -C test/files ./test_app
+            gtar -czvf test/files/pkg-app.spl -C test/files/test_app .
```
acharlieh: This change to the packaging is incorrect. A splunk app is packaged so that the directory the app needs to be installed is the root of the tarball. 

e.g. an app with the url test_app would have all entries in the tarball start test_app/

alecxvs: I think you're right, but I can't fathom how I revisited this two or three times and landed on the wrong thing. Maybe it's OSX making it impossible to tell the difference, since both this app and a properly packaged app extract to similar outputs.

libraries/conf_helpers.rb
```diff
@@ -46,7 +46,8 @@ def self.parse_config(conf_body)
         case ln
         when /^\s*\[([^#]+)\]\s*/
           current_section = Regexp.last_match[1]
-        when /^\s*([^#]+?)\s*=\s*([^#]+?)\s*(?:#.*)?$/
+          current_config[current_section] ||= {}
+        when /^\s*([^#]+?)\s*=\s*([^#\v]*?)\s*(?:#.*)?$/
```
acharlieh: I'm curious what lead to this change in the second capturing group, to add a negative match on vertical tab?

(We probably should add line continuation parsing as well... I wonder if we need an actual parser generator...)...

alecxvs: It was capturing the newline in the values, which lead to weird things like "someKey = \n". I added some unit tests for this case.

.kitchen.yml
alecxvs: 87e832c

resources/splunk_app.rb
```diff
@@ -162,30 +163,40 @@ class PackagedApp < SplunkApp
 
   action :install do
     return unless !version || changed?(:version)
+    @converge_actions = nil # Clear resource updates by property change
```
gravesb: Can you provide a bit more context on what this change is for? The existing comment doesn't really help me much.

resources/splunk_app.rb
```diff
@@ -162,30 +163,40 @@ class PackagedApp < SplunkApp
...
+      path app_path.to_s
+      recursive true
       action :create
     end
 
     backup_app if changed? :version
 
-    poise_archive package_path.to_s do
-      destination((app_cache_path + 'new').to_s)
+    temp_path = (app_cache_path + 'new').to_s
```
gravesb: I'm curious, what issue did moving this out to a separate variable declaration fix?

alecxvs: I couldn't determine why I did this
d796a9024a997db7d3199601c20328fa1a0b28e3

It seems to work with the integration tests I'm able to run. (Someone else will have to run integration tests, because I can't get the disk space to do so)

resources/splunk_app.rb
alecxvs: It's a little difficult... So with resources, they're given a desired property and I'm populating the existing property. In the case this is necessary, there is a 1.0.0-somesnapshotversion on disk, and the resource is given `version 1.0.0`. The app it downloads will have 1.0.0-somesnapshotversion which is the exact same version, and I do checks for this later in the resource to determine whether or not we need to upgrade the app (which would be a change in the resource, triggering notifications).

The problem is that Chef considers a difference in the given property and the existing property to be a "change", reports this in the output, and adds a "converge action" to this list. Since a difference in given properties isn't indicative of an actual change, I'm clearing the converge actions before doing anything else.

gravesb: I would be fine putting something along those lines into the code comment.  It's a big comment but I think it's warranted here.

alecxvs: f9b4a4f05d02cbeb299c233674f24bf21af2affc

gravesb (approved):

libraries/conf_helpers.rb
acharlieh: Ah wait... it's not the vertical tab addition it's the change of `[^#]+?` to `[^#]*?` which is doing the magic here. as the former requires at least 1 character, as opposed to the latter which requires no characters

libraries/provider_helpers.rb
```diff
@@ -80,7 +96,7 @@ def validate_versions # rubocop:disable Metrics/PerceivedComplexity, Metrics/Cyc
         end
 
         # Check that the package's version is not a pre-release when we really expect a release
-        if !app_version.prerelease? && pkg_version.prerelease?
+        if current_resource.version && !current_resource.version.prerelease? && pkg_version.prerelease?
```
acharlieh: For my clarity... `current_resource.version` = installed version.... `app_version` = desired version ... and `pkg_version`  = the version that is exactly installed...

I'm trying to remember but I don't know that this check is actually valid... 

going from "1.0.0" -> "2.0.0-alpha" is totally valid... going "1.0.0" -> "1.0.0-alpha" is not valid, but if 1.0.0 is installed, and 1.0.0 is the desired version, we shouldn't be downloading the package anyways...

resources/splunk_service.rb
```diff
@@ -81,7 +81,7 @@ def service_action(desired_action)
     def initialize_service
       return unless CernerSplunk::PathHelpers.ftr_pathname(install_dir).exist?
 
-      cmd = "#{command_prefix} enable boot-start#{platform_family?('windows') ? '' : " -user #{current_owner}"} --accept-license --no-prompt"
+      cmd = "#{command_prefix} enable boot-start#{platform_family?('windows') ? '' : " -user #{current_owner}"} --accept-license --answer-yes --no-prompt"
```
acharlieh: If I'm agreeing to the license on Windows (per resources/splunk_install.rb), would this code ever be hit in a non-windows context?

Also does this perform the ftr steps as the correct user?

libraries/provider_helpers.rb
alecxvs: 05dd3b618685d1d409e9730a84d78322933ca46f

resources/splunk_service.rb
alecxvs: Accepting the license shouldn't execute FTR
FTR executes as the boot-start user

libraries/provider_helpers.rb
```diff
@@ -69,21 +85,23 @@ def validate_extracted_app
...
+        unless new_cache_path.exist?
+          raise "Downloaded app is not structured correctly. Expected sub-directory named #{name}, got #{(app_cache_path + 'new').children(false)}"
+        end
+
         app_version = version
         pkg_app_conf = CernerSplunk::ConfHelpers.read_config(new_cache_path + 'default/app.conf')
-        return true unless app_version && pkg_app_conf.key?('launcher')
+
+        # If the version is not specified, or the version does not exist in the currently installed app, then continue to install.
+        return true unless app_version && pkg_app_conf.key?('launcher') && pkg_app_conf.dig('launcher', 'version')
```
acharlieh: This doesn't line up with the comment... This is checking the desired version and the packaged app version... not the installed app version...

alecxvs: There were actually a few errors in this validation...
84ed597deb78a3d626dfa879b5d2f40bca2c1ebe

I added a check to validate_packaged_app to ensure the downloaded app has a version, removed a redundant and untested check for the downloaded app directory (validate_packaged_app already does this), and changed this to check that the current desired version and current installed version exist

resources/splunk_app.rb
```diff
@@ -135,11 +130,12 @@ class CustomApp < SplunkApp
   resource_name :splunk_app_custom
 
   action :install do
-    return unless !version || changed?(:version)
+    return unless !new_resource.version || changed?(:version)
 
     directory app_path.to_s do
       user current_owner
       group current_owner
+      recursive true
```
acharlieh: Out of curiosity... what issues were you seeing here? 

I think the behavior of `recursive true` can be a bit surprising if not expected in that it at least used to be that created intermediate directories have the default user/group owners... only the end directory has the specified user/group owner.

acharlieh (approved):

resources/splunk_app.rb
alecxvs: So I'm sure I've encountered some case where the parent directory was missing, but I can't figure out what that case is now since all of the splunk standard directories are provided by the installation. I'm inclined to leave this here because of uncertainty, but also to smooth over the use case of specifying a different base directory for the app.

